PR #14, opened 2020-12-08
1.2 Grammar of graphics

joelostblom:
I started doing what we said during the meeting, but I got too annoyed copyting all code chunks around, inserting and resizing images, so I spent some more time on the knitr hook this morning and got it working!

The knitr hook uses vegawidget to create SVGs automatically from Altair chunks and can easily be changed to output HTML if Ines fixes the framework bug (I tested and it works with regular RMarkdown to create interactive plots). This workaround is a bit hacky since it relies on string manipulation of the python code chunk, but I think it is the best we can do until RStudio and Reticulate officially supports Altair. I can't immediately think of anything in my lectures that would break this, so should be good moving forward (it relies on that the last part of the python code chunk is an Altair plot, which I believe it should always be).

Also a reminder that the previous PR is ready for re-review.


ttimbers:
Ha! You are indeed persistent!!! Nice work on this! I will review tomorrow afternoon!

joelostblom:
Force pushed to rebase on the latest changes in the parent PR

chapters/en/slides/module1/module1_02.Rmd
```diff
@@ -0,0 +1,280 @@
...
+
+## A high level grammar of graphics helps us compose plots effectively
+
+- Simple grammatical components combine to create visualizations.
+- The Altair visualization grammar consist of three main components:
+    1. Create a chart.
+    2. Add a graphical mark.
+    3. Encode dataframe columns as visual channels.
+- In code, it looks like this: `Chart(data).mark().encode(x, y, ...)`.
+
```
ttimbers, line 95: This slide needs a visualization! We are at slide 4 in a module of a viz course with no viz! I would have a simple scatter plot actually generated by altair, and then draw using illustration software some arrows/circles/whatever on top of it pointing out what is the chart, what is a mark, etc

chapters/en/slides/module1/module1_02.Rmd
```diff
@@ -0,0 +1,280 @@
...
+cars = data.cars()
+cars
+```
+
+Notes:
+When specifying data in Altair,
+it is important that dataframes are in the ["tidy"](http://vita.had.co.nz/papers/tidy-data.html) format.
+They should consist of a set of named data *columns* (also called *fields* or *variables*)
+with one feature each
```
ttimbers, line 129: Here we have just used 3 names in one sentence to describe a columnar variable. There is already a lot of jargon in this course, let's stick with one name for this. I would vote for variable.

chapters/en/slides/module1/module1_02.Rmd
```diff
@@ -0,0 +1,280 @@
...
+we will often use data from the [vega-datasets](https://github.com/vega/vega-datasets) package.
+Many of these datasets are directly available as Pandas dataframes,
+such as the `cars` dataset.
+
+---
+
+## Let's create our first plot
+
+```{python}
+import altair as alt
```
ttimbers, line 141: Let's talk about this in the transcript. It might be a while since their programming in Python course, so a reminder what this is doing here, and that altair is the name of the package we are importing and alt is just the nickname/alias would probably be welcome by the learners.

chapters/en/slides/module1/module1_02.Rmd
```diff
@@ -0,0 +1,280 @@
...
+
+alt.Chart(cars).mark_point()
```
ttimbers, line 143: Should we first show an empty chart, i.e., call `alt.Chart(cars)`, and then show adding a mark?

chapters/en/slides/module1/module1_02.Rmd
```diff
@@ -0,0 +1,280 @@
...
+## Mapping a data feature to the y-scale
+
+```{python}
+alt.Chart(cars).mark_point().encode(
+    x='Miles_per_Gallon',
+    y='Horsepower')
+```
+
+Notes:
+Though we've now separated the data by one attribute,
```
ttimbers, line 199: In the last slide we referred the this as a column, now here its an attribute. Let's pick a name and consistently use it.

chapters/en/slides/module1/module1_02.Rmd
```diff
@@ -0,0 +1,280 @@
...
+let's code!
+
+---
+
+## Sample data can be found in Altair's companion package vega_datasets
+
+```{python}
+from vega_datasets import data
+
+cars = data.cars()
```
ttimbers, line 121: I don't mean to be negative, however, could we pick a more interesting data set? We need to get folks excited and motivated to stick around! Maybe a pre-amble also of a question we'd like to answer, and then we introduce the altair code slowly as you do now to answer it?

In online courses it is important to motivate, or we lose learners. There's not the same level of engagement. You could borrow a question and viz from Introduction to Data Science? See here: https://ubc-dsci.github.io/introduction-to-datascience/viz.html#choosing-the-visualization

exercises/en/exc_01_03.py
```diff
@@ -1,9 +1,10 @@
...
 
-print(candybar_df.head())
-
-candybar_dim = candybar_df.____
-print(candybar_dim)
\ No newline at end of file
+alt.Chart(___).mark___.encode(
+    x=___,
+    y=___,
+    color=___)
```
ttimbers, line 10: I need to see the data frame to complete the question successfully. I can't remember the column names. Also, I think somewhere we need to explicitly tell people that the column names need to be quoted (for poor R users like me).

ttimbers (changes requested):
Good first draft! Some adjustments to make but you are on the right track!

chapters/en/slides/module1/module1_02.Rmd
```diff
@@ -0,0 +1,280 @@
...
+rather than having to specify *how* to implement the visualization in terms of for-loops, low-level drawing commands, *etc*.
+For example,
+you would say "color my data by the column 'country'"
+instead of "go through this data frame and plot any observations of country1 in blue, any observations of country in red, etc".
+
+Declarative visualization tools lets you think about **data and relationship**,
+rather than **plot construction details**.
+A key idea is that you declare links between data fields and visual encoding channels,
+such as the x-axis, y-axis, color, *etc*.
+The rest of the plot details are handled automatically.
```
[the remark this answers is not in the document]
joelostblom: I have changed this to have 2 slides for each, one with descriptional bullets points and oen with pseudocode and a plot. I have a final sentence comparing the two, but not a separate slide.

chapters/en/slides/module1/module1_02.Rmd
```diff
@@ -0,0 +1,280 @@
...
+
+- Focus on data and relationships.
+    - Often includes data, graphical marks, and encoding channels.
+- Specify *what* should be done
+    - "Assign colors based on the values in column A"
+
+Notes:
+By *declarative*,
+we mean that you can provide a high-level specification of *what* you want the visualization to include,
+in terms of *data*, *graphical marks*, and *encoding channels*,
```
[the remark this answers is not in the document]
joelostblom: Done

chapters/en/slides/module1/module1_02.Rmd
joelostblom: I have added two visualizations before this points so that slides feel less text heavy.

chapters/en/slides/module1/module1_02.Rmd
joelostblom: I have clarified this to:

> ...
> and a set of named data *columns*
> with one feature each
> (you might also have heard these called *fields* or *variables*,
> but we will stick to *columns* for clarity).

I stuck to "columns" because I think it is clearer since we are referring to the other axis as "rows", but we can discuss more. But in general I think we should use the same as in the prerequisite MCL course, @hfboyce which nomenclature did you use there?

chapters/en/slides/module1/module1_02.Rmd
joelostblom: Done

chapters/en/slides/module1/module1_02.Rmd
joelostblom: I agree this makes sense in general, but the mark is a required property of Altair/vega chart so leaving it out raises an error.

chapters/en/slides/module1/module1_02.Rmd
joelostblom: Done

chapters/en/slides/module1/module1_02.Rmd
joelostblom: I kept the car data because of time concerns and that I want to show `vega_datasets` but I have written more to build it up similar to what you do in the link above including a questions we might have of the data so that it hopefully feels more interesting.

exercises/en/exc_01_03.py
joelostblom: This is available in the hints section, do you think that is enough?

chapters/en/slides/module1/module1_02.Rmd
```diff
@@ -0,0 +1,280 @@
...
+rather than **plot construction details**.
+A key idea is that you declare links between data fields and visual encoding channels,
+such as the x-axis, y-axis, color, *etc*.
+The rest of the plot details are handled automatically.
+
+---
+
+## The Python plotting landscape
+
+<img src="/module1/py-plotting-landscape.png" alt="The Python plotting landscape" width="100%"></img>
```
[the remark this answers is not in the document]
joelostblom: I described this in the transcript for now in the interest of time and to better motivate why we chose Altair.

joelostblom:
@ttimbers Although I merged this, let me know if you think any of your comments is unresolved and I will open and issue to track that. Sorry for replying so late, I realize you might not recall the details of your comments since they were made several weeks ago.

chapters/en/slides/module1/module1_02.Rmd
joelostblom: Writing the faceting and EDA chapters, it would be nice to use "variable" since there now already is something called "column" in the plot grid. @hfboyce Can I change this? Or what did you use previously?

hfboyce: In programing in python they should be comfortable with variable or column. We used both interchangeably in that course. In the ML we used "feature" as well.

PR #16, opened 2020-12-08
1.3 lines and layers

joelostblom:
This is builds upon #14 so merge that first.
I have set the base branch to be 14 here so that it is easy to view only these changes.
It will change to main automatically when that PR is merged.

I wrote a couple of simple test using only the string,
but if I understood correctly some of the MDS students
will write the more complex one working with the actual object,
or should I do 1-2 of them too as a test?

There is a little bit of clean up in this PR also
since some of it dependent on the content changes I made here.
I tried to keep it minimal so it hopefully didn't become confusing.


joelostblom:
Force pushed to rebase on the latest changes in the parent PR

chapters/en/slides/module1/module1_03.Rmd
```diff
@@ -0,0 +1,253 @@
...
+from vega_datasets import data
+
+cars = data.cars()
+
+alt.Chart(cars).mark_point().encode(
+    x='Year',
+    y='Miles_per_Gallon')
+```
+
+Notes:
```
ttimbers, line 46: Again, we need to motivate with a compelling case and story.

ttimbers (changes requested):
I have a big picture concern here - I think two important concepts are being taught at once and it conflates things. One concept is using a line as a mark (why and how), as well as data aggregation (why and how).

I suggest we split this slide deck into two. The first being a simpler case, where we have 1-3 lines we'd like to visualize. The second being aggregation (at least for line charts).

chapters/en/slides/module1/module1_03.Rmd
joelostblom: I have rewritten to have more of a narrative throughout this slide deck.

joelostblom:
> I have a big picture concern here - I think two important concepts are being taught at once and it conflates things. One concept is using a line as a mark (why and how), as well as data aggregation (why and how).
> 
> I suggest we split this slide deck into two. The first being a simpler case, where we have 1-3 lines we'd like to visualize. The second being aggregation (at least for line charts).

@ttimbers I agree with the big picture concern, I have separated the sections clearer and teach aggregations with points before introducing the lines and then finally the layers. We can split this into different slides decks if you still think it is easier, but it would be just a few slides each. I don't have a preference, leaving as is for now.

PR #22, opened 2021-01-07
404 build issue

joelostblom:
Making this a PR instead of pushing on main to make sure it fixes the build error on netlify (works locally). Will merge if tests pass.


joelostblom:
Merging, not sure why this was needed as the previous 404 page was the default one used in the Gatsby repo and this one is a custom smaller version. Anyways, it works now.

PR #23, opened 2021-01-07
2.1 visual channels intro

joelostblom:
This is ready for review but it might makes sense to wait until the entire module is done so that you can chunk it all at once.


chapters/en/slides/module2/module2_01.Rmd
```diff
@@ -0,0 +1,234 @@
...
+In this lecture,
+we will see how we can also use areas and bars for this purpose.
+Before we dive in to the code,
+let's discuss how different visual channels,
+such as position, area, etc,
+can impact how easy it is for us
+to accurately interpret the plotted data.
+
+As in many cases,
+an efficient way to learn this topic is through personal experience.
```
ttimbers, line 31: Do we mean personal experience here? Or do we mean practice?

chapters/en/slides/module2/module2_01.Rmd
```diff
@@ -0,0 +1,234 @@
...
+
+Notes:
+In the image in this slide,
+try to estimate how many times larger the big circle is compared to the small one.
+
+---
+
+## Comparing sizes is easier for some geometrical objects than for others
+
+![](/module2/bar-comparison.png)
```
ttimbers, line 49: This bar and square combo does not make it easy to answer this question... Could we do one where it's obviously a quarter? And then, the circle one should be a quarter? I think the contrast would be quite stark.

ttimbers: I see that these slides are borrowed, but I think we could easily remake our own and still attribute where the inspiration came from.

chapters/en/slides/module2/module2_01.Rmd
```diff
@@ -0,0 +1,234 @@
...
+but also when creating plots for yourself.
+These two examples are originally from [Jeffrey Heer's PyData talk](https://www.youtube.com/watch?v=hsfWtPH2kDg),
+who is visualization researcher at the university of Washington
+and whose research group created D3 and VegaLite
+(which Altair is based on).
+
+---
+
+## Summary of visual channel efficiency
+
```
ttimbers, line 83: This slide should have a header/footer clearly stating that these numbers have a meaning, and what that meaning is (ranking of ease for humans to distinguish). Some folks may just read the slides...

chapters/en/slides/module2/module2_01.md
```diff
@@ -0,0 +1,184 @@
...
+easiest for humans to decode. Luckily, there has been plenty of research
+in this area, which can be summarizes in the schematic on this slide.
+
+Position is by far the best and therefore we should put our most
+important comparison there. Using position often means that we can’t use
+other things such as length or angle (like the angle in a pie chart),
+but we can add size or color to represent other relationships. Even if
+it is hard to tell exact information from these (is this color/dot 2x
+darker/bigger than another?) they are good to give and idea of trends in
+the data (are these dots darker/bigger than the others?
```
ttimbers, line 74: You can combine position and angle (think wind direction visualization). I think a pie chart is more of an area thing. I think this second paragraph needs some work to make it clear and effective.

chapters/en/slides/module2/module2_01.md
```diff
@@ -0,0 +1,184 @@
...
+## Don’t use 3D without a good reason
+
+### Unnecessary 3D makes plot interpretation harder
+
+![](/module2/bad-3d-barchart.png)
+
+Notes: The biggest issues with using 3D is when it is used unnecessarily
+(like a 3D bar or pie chart), as the only way to compare position (like
+a 3D scatter plot), and when they are represented on a 2D medium like a
+paper where they can’t be rotated. In this slide we see a 3D bar chart,
```
ttimbers, line 87: I think the bigger issue is not that they cannot be rotated, but that the extra cognitive load in reading them takes away from the pop out of differences.  The camera angle also is a thing, I agree.

chapters/en/slides/module2/module2_01.md
```diff
@@ -0,0 +1,184 @@
...
+---
+
+## Meaningful 3D can facilitate plot interpretation
+
+![](https://media.github.ubc.ca/user/1751/files/da4e1b00-0fd3-11eb-940d-0b06a28a0186)
+
+Notes: Sometimes 3D can be useful, like a topographical map or a protein
+folding visualization. But be cautious, we saw above that even in
+naturally 3D systems like blood vessels it is still mentally more
+complex when they are 3D. [Claus Wilke’s has a good
```
ttimbers, line 102: I don't think the naturally 3D system you are referring to here.

chapters/en/slides/module2/module2_01.md
```diff
@@ -0,0 +1,184 @@
...
+additional 2D plot.
+
+---
+
+## Properly designed visualizations help saving lives
+
+![](/module2/blood-vessels-40.png)
+
+Notes: How much these best practices actually matters might be a bit
+abstract until you gain personal experience from it, Therefore, I want
```
ttimbers, line 120: Again, not sure personal experience is the right term here either (neither is practice...)

chapters/en/slides/module2/module2_01.md
```diff
@@ -0,0 +1,184 @@
...
+shear stress areas were now correctly identified.
+
+Overall, these two tweaks more than doubled the outcome accuracy, from
+40% to 90%. A huge increase from modification that might have seemed to
+be a mere matter of taste unless you knew visualization theory! So, if
+anyone tells you that visualization of data is not as important as other
+components, you can tell them about this study and ask them what kind of
+visualization they want their doctor to look at when analysing their
+arteries.
+
```
ttimbers, line 179: Really like this story!

ttimbers (changes requested):
Please see comments to improve this for the next draft. But good start!

chapters/en/slides/module2/module2_01.md
joelostblom: Good point, I guess wind/vector field has ange, position, and also length. Such a rich viz, I should try use it more often! 

Regarding pie charts, they seem to be classified as "angle" in the viz studies I have see references to ([cited here](https://socviz.co/lookatdata.html#visual-tasks-and-decoding-graphs)), e.g. in a study from the D3 author with this schematic:

image
 
But I agree with you that there is an area interpretation as wel, so not sure which is dominant. For this slide it would make sense to ref the tiling line as in the picture instead for angle. I will clarify and also make more effective overall.

chapters/en/slides/module2/module2_01.Rmd
joelostblom: I have made the slides myself, but the idea is from the viz research talk I referenced and I used the same proportions they used since I thought there might be a reason behind it. I can remake them using 1/4 instead, it might be too easy to interpret the circle correctly then? Let's see.

joelostblom (commented):
Thanks for you comments @ttimbers !

Issue #28, opened 2021-01-15
2.2 Make reticulate work with >5k rows of data and fix html table length


joelostblom:
1. Since Altair includes all data in the json plot spec, it raises an error when there are more than 5000 rows in the data to avoid creating massive files. The ideal workaround for this is to use Altair's `data-saver` transformer, but reticulate freezes when I try this (I think because it involves another Python instance and I believe I read that reticulate is limited to just one Python instance for some reason). 
2. Another workaround is disabling the warning, this works but reticulate is so slow to create the plots (several minutes per plot) that it is unusable (and we would probably not want those big md-files on github anyways).
3. A third way is to use the json transformer that create a local temporary json file, but with this approach `vegawidget` fails to create the svg. I have opened an issue about this.
4. I tried creating a png instead via vegawidget, but apparently the png converter requires svg creation first so same problem as above.
5. My current workaround is to filter the gapminder dataset to the 10 biggest countries, so that it is only 2k rows. Although the trends are similar as with the entire dataset, several of the points I make in the end slides are about the world population so it is a bit unfortunate that we have to resort to this.

joelostblom:
Fixed with html figures

PR #38, opened 2021-01-21
2.2 Area charts

joelostblom:
Ready for review!

There are still a few small things to fix here, but they depend on the resolution of a new issue I ran into working with reticulate.[ I  opened an issue to track it](https://github.com/UBC-MDS/exploratory-data-viz/issues/28), so you can review this PR and merge when ready and we will solve the outstanding issue later. In essence we can't work with datasets with more than 5k rows and the normal ways around it are not supported by reticulate.


chapters/en/slides/module2/module2_02.Rmd
```diff
@@ -0,0 +1,279 @@
...
+| income_group          | Income group [as specified by the world bank in 2018]                                        |
+| life_expectancy       | The mean number of years a newborn would <br>live if mortality patterns remained constant    |
+| income                | GDP per capita (in USD) <em>adjusted <br>for differences in purchasing power</em>            |
+| children_per_woman    | Average number of children born per woman                                                    |
+| child_mortality       | Deaths of children under 5 years <break>of age per 1000 live births                          |
+| pop_density           | Average number of people per km<sup>2</sup>                                                  |
+| co2_per_capita        | CO2 emissions from fossil fuels (tonnes per capita)                                          |
+| years_in_school_men   | Mean number of years in primary, secondary,<br>and tertiary school for 25-36 years old men   |
+| years_in_school_women | Mean number of years in primary, secondary,<br>and tertiary school for 25-36 years old women |
```
joelostblom, line 45: This table is too long but it can be fixed by removing columns if we don't find a reticulate workaround. We can probably remove column anyways and maybe copy this data to a UBC repo (maybe this one?).

chapters/en/slides/module2/module2_02.Rmd
```diff
@@ -0,0 +1,279 @@
...
+
+## Reading in the data with pandas
+
+```{python}
+import pandas as pd
+
+
+url = 'https://raw.githubusercontent.com/UofTCoders/workshops-dc-py/master/data/processed/world-data-gapminder.csv'
+gm = pd.read_csv(url, parse_dates=['year'])
+gm = gm[gm['country'].isin(gm[gm['year'] == '2018'].nlargest(10, ['population'])['country'])]
```
joelostblom, line 67: Filter data because reticulate doesn't support the Altair workaround for over 5k rows in the dataset.

chapters/en/slides/module2/module2_02.Rmd
```diff
@@ -0,0 +1,279 @@
...
+of the visualization principles
+with data we already know.
+
+## Reading in the data with pandas
+
+```{python}
+import pandas as pd
+
+
+url = 'https://raw.githubusercontent.com/UofTCoders/workshops-dc-py/master/data/processed/world-data-gapminder.csv'
```
joelostblom, line 65: Should we move this data into our repo?

chapters/en/slides/module2/module2_02.Rmd
```diff
@@ -0,0 +1,279 @@
...
+so it is easy to see how the total world population has changed over time.
+
+From the stacked colored regions,
+we also get a rough idea of how each region has grown,
+but it is hard to compare exactly,
+especially for regions that are not stacked next to each other.
+For example,
+we can't really tell if Europe, Africa, or the Americas
+has the largest population in the most recent year.
+
```
ttimbers, line 198: Add a summary sentence here with that take home for what this plot is good for, and when to use it (ending on what its problems are makes it seem likes it not a great plot type).

chapters/en/slides/module2/module2_02.Rmd
```diff
@@ -0,0 +1,279 @@
...
+here it is immediately clear that Africa has a bigger population 
+than Europe and the America,
+and we can see that the populations shifted around year 2000.
+
+On the other hand,
+it's quite cognitively demanding
+to try to reconstruct the total world population
+by adding all the lines up together,
+especially over time!
+
```
ttimbers, line 225: Add a summary sentence here with that take home for what this plot is good for, and when to use it (ending on what its problems are makes it seem likes it not a great plot type).

chapters/en/slides/module2/module2_02.Rmd
```diff
@@ -0,0 +1,279 @@
...
+However,
+when visualizing the trends over time for multiple groups,
+lines and areas have different advantages,
+as we will see next.
+
+---
+
+## Area charts are preferred when the total of the groups is the most important
+
+```{python}
```
ttimbers, line 179: Should the y-axis be something else here? Like cumulative sum? It's a bit different than how the data is interpreted on the next slide, but the axis labels are the same.

chapters/en/slides/module2/module2_02.Rmd
```diff
@@ -0,0 +1,279 @@
...
+Fortunately,
+the world population is predicted
+to stabilize around 11 billion people at the end of the century.
+
+Why?
+Because as living conditions improve around the world,
+child mortality is going down,
+which means that the number of children per women is decreasing
+towards a level than means that the current population will be sustained
+rather than increase.
```
ttimbers, line 268: This is written as fact, where it is actually a researchers hypothesis. We should make sure that is clear. 

Also, this sentence is a run on does not read right: "Because as living conditions improve around the world,
child mortality is going down, which means that the number of children per women is decreasing towards a level than means that the current population will be sustained rather than increase."

Decreases in child mortality is not causing the number of children per women to decrease...

ttimbers (changes requested):
Please see and address my review comments. Also, this is just one type of area chart - I assume other types will follow?

chapters/en/slides/module2/module2_02.Rmd
joelostblom: My understanding is that decreasing child mortality is one of the main drivers in reducing the number of children per woman. This is presented as a causative relationship by both gapminder and [Our world in data](https://ourworldindata.org/fertility-rate#increasing-well-being-and-status-of-children) with several academic references, and I presented it the same way since these organizations are global health experts.

There are of course other factors that contribute to reduced fertility rates, and this relationship is not causative in the same strong sense as in natural sciences ([for example not in Niger](https://www-nature-com.myaccess.library.utoronto.ca/articles/542414a)), but this is a limitation of any observational study (so most global health science). I can rephrase to reflect this, but I still the most accurate to portray this as the literature in the field does, maybe something like:

"In countries where child mortality is high, women will often give birth to more children than they want in order to protect against losing some of their children prematurely. Lower child mortality leads to more predictable family planning and as child mortality goes down, the number of children per woman approaches 2 in the majority of countries. "

chapters/en/slides/module2/module2_02.Rmd
joelostblom: Added a todo to revisit later

PR #42, opened 2021-01-22
2.4 faceting

joelostblom:


chapters/en/slides/module2/module2_04.Rmd
```diff
@@ -0,0 +1,143 @@
...
+
+```{r setup, include=FALSE}
+source('../../../../slide-setup.R')
+# Remember to also manually update the YAML title above
+```
+
+type: slides
+
+# Creating subplots via faceting
+
```
ttimbers, line 19: Can you add a slide before this to set this up? Just a reminder of the question, we asked and answered with this histogram, without colouring by region. And then we can formulate a new question that would be answered by what is done on this slide here.

chapters/en/slides/module2/module2_04.Rmd
```diff
@@ -0,0 +1,143 @@
...
+    y='count()',
+    color='region'
+).properties(width=200, height=200
+).facet('region')
+```
+
+Notes:
+Placing color segments next to each other
+would make it easier to compare their heights.
+Altair does not yet support this type of plot,
```
ttimbers, line 71: Wouldn't this also be problematic when we get too many categories? That is another reason to facet.

chapters/en/slides/module2/module2_04.Rmd
```diff
@@ -0,0 +1,143 @@
...
+
+Now we see that the bulk of the distribution for Europe
+is to the right at higher life expectancies
+whereas it is lower for African countries.
+
+---
+
+## Subplots are often laid out in even grids
+
+```{python}
```
ttimbers, line 88: What is with the empty plot? Here and on the next slide? We at least need to comment on it. If possible, could we show them how to remove it? It would be a more effective plot with it removed.

chapters/en/slides/module2/module2_04.Rmd
```diff
@@ -0,0 +1,143 @@
...
+## Faceting creates subplots for different subsets of the data
+
+```{python}
+alt.Chart(gm2018).mark_bar().encode(
+    x=alt.X('life_expectancy', bin=alt.Bin(maxbins=30)),
+    y='count()',
+    color='region'
+).properties(width=200, height=200
+).facet('region')
+```
```
ttimbers, line 66: I know they might not fit the slide, but wouldn't these be better to compare if we stacked them vertically? If it's super difficult on the slide, or looks funny, we might want to mention it at least, and then show that on another slide, where we just focus on that. That seems like missing material a bit... When to stack vertically, horizontally, or in a grid...

chapters/en/slides/module2/module2_04.Rmd
```diff
@@ -0,0 +1,143 @@
...
+## Faceting across two categorical variables
+
+```{python}
+alt.Chart(gm2018[gm2018['income_group'].isin(['High', 'Low'])]).mark_bar().encode(
+    x=alt.X('life_expectancy', bin=alt.Bin(maxbins=30)),
+    y='count()',
+    color='region'
+).properties(width=200, height=140
+).facet(column='region', row='income_group')
+```
```
ttimbers, line 119: We should comment that these plots are empty for a different reason.

chapters/en/slides/module2/module2_04.Rmd
```diff
@@ -0,0 +1,143 @@
...
+(we have filtered the data to save space on the slide).
+
+Depending on the exact relationship you want to investigate,
+you can choose which dataframe columns
+you encode in which visual channels
+according to the principles we have discussed
+to best bring out the comparison of interest in the visualization.
+
+---
+
```
ttimbers, line 140: Can we add a wrap-up slide comparing the plot we started with to one of the more effective plots side-by-side and reiterate the power of facetting?

ttimbers (changes requested):
Feedback for chapter 4.

chapters/en/slides/module2/module2_04.Rmd
```diff
@@ -0,0 +1,143 @@
...
+source('../../../../slide-setup.R')
+# Remember to also manually update the YAML title above
+```
+
+type: slides
+
+# Creating subplots via faceting
+
+Notes:
+Exploring mutiple vars
```
joelostblom, line 21: TODO

chapters/en/slides/module2/module2_04.Rmd
joelostblom: I don't have any directly good guideline for when to stack in different ways, this would be interesting to discuss! I also have found a way to modify slight height, not sure if it will work for an indivisual slide but could try. I will ping you when  I revisit this.

chapters/en/slides/module2/module2_04.Rmd
joelostblom: It's the default behavior, not sure how easy it would be to remove for just one subplot, and I am not sure if it would be better if removed, I think aligning to the top would be ideal. I opened https://github.com/vega/vega-lite/issues/7194

chapters/en/slides/module2/module2_04.Rmd
joelostblom: The reason for this empty plot is similar, "we ran out of plot", just that now they are at specific combinations of rows and columns, not necessarily at the end. Will elaborate.

chapters/en/slides/module2/module2_04.Rmd
joelostblom: Added a comment about this and a few other slides.

chapters/en/slides/module2/module2_04.Rmd
joelostblom: Added a comment about this for when color is the only channel used to distinguish groups

PR #44, opened 2021-01-23
3.1 Single column distributions

joelostblom:
@ttimbers I ended up finishing the two first decks of this module. They are a bit rougher than the last PRs but ready to review. I will add the last 1-3 decks next week, recording my thoughts here about how I am thinking about the potential remaining ones.

1. For sure include comparing multiple columns (why bar plots are not good, box plots, violins)
2. Probably include 2D histograms for when scatter plots become saturated (I didn't do it this way in 531, but it does fit naturally here).
3. Maybe include the distribution equivalent of categoricals (so counting in bars and heatmaps).
    - I think this fits thematically in module 3, but it could also go in the module 4 (EDA), which would make these two more balanced content wise.


joelostblom:
Hmm for some reason we didn't get a deploy preview here, remaking the PR.

PR #45, opened 2021-01-23
3.1 Single column distributions

joelostblom:
@ttimbers I ended up finishing the two first decks of this module. They are a bit rougher than the last PRs but ready to review. I will add the last 1-3 decks next week, recording my thoughts here about how I am thinking about the potential remaining ones.

1. For sure include comparing multiple columns (why bar plots are not good, box plots, violins)
2. Probably include 2D histograms for when scatter plots become saturated (I didn't do it this way in 531, but it does fit naturally here).
3. Maybe include the distribution equivalent of categoricals (so counting in bars and heatmaps).
    - I think this fits thematically in module 3, but it could also go in the module 4 (EDA), which would make these two more balanced content wise.


chapters/en/slides/module3/module3_01.Rmd
```diff
@@ -0,0 +1,274 @@
...
+
+Before diving into these comparisons,
+we want to understand the data for each of the columns in our dataset.
+
+Let's start by looking at the ratings the movies received
+and explore the best ways to visualize this data.
+
+---
+
+## Visualizing a single column with a point plot along one axis
```
ttimbers, line 61: This is spelled incorrect on the rendered slides deployed by netlify, there it reads: "ReprestnData points from a single variable", not sure why the mismatch...

chapters/en/slides/module3/module3_01.Rmd
```diff
@@ -0,0 +1,274 @@
...
+alt.Chart(movies[:5]).mark_point().encode(
+    x='runtime')
+```
+
+Notes:
+One of the simplest visualization we could perform
+with a single dataframe column
+is to plot each individual observations
+as a point.
+
```
ttimbers, line 76:
```suggestion
Visualizing data distributions are useful in understanding the measurements
of a variable, as doing this allows us to quickly and easily see the range and
frequency of values that occur.

One of the simplest visualizations to understand a 
distribution of a single variable (i.e., a single column from a tidy data frame) 
is to plot each individual observations as a point.

```

chapters/en/slides/module3/module3_01.Rmd
```diff
@@ -0,0 +1,274 @@
...
+```
+
+Notes:
+Instead of visualizing each individual observation,
+we can use the histograms we learnt about in last module,
+where we divide one axis into groups (or "bins")
+and represent the count of observations in each group with bar.
+
+In this plot,
+we can directly see that most movies are around 100 min in length.
```
ttimbers, line 161: Add comments on range of the data - what is the minimum and maximum movie length. This is something visualizing distributions lets us know. Also, what about the shape? Are the values uniformly distributed? Or do most values fall in one peak in the middle? Also say why you can say that most movies are around 100 min in length (because there is a single peak, meaning most values in the data set fall there, and that peak is centered around 100).

Then take a breath and start discussing multiple distributions.

chapters/en/slides/module3/module3_01.Rmd
```diff
@@ -0,0 +1,274 @@
...
+```
+
+Notes:
+Let's start by comparing the movie length broken down per country.
+Here we use faceting which we learnt about in the last module,
+to create one histogram per country.
+
+You can see that this dataset only contains movies made in the UK or the US.
+It does seem like the distribution of movies is pretty equal between the two countries,
+but it is a bit hard to see because there are so many more movies made in the US.
```
ttimbers, line 187: I don't know if I agree. The shape looks flatter (because of the binning), so maybe I would say here that the range and the centre looks similar, but maybe the shape is different, its hard to tell because there are so many more movies in the US...

chapters/en/slides/module3/module3_01.Rmd
```diff
@@ -0,0 +1,274 @@
...
+and the total number of observations.
+
+If we want to drill down and compare the distribution shape more closely
+regardless of the number of observations
+we could define the y-axes to be independent for each subplot,
+instead of linked to the same axis range.
+
+Now we can see that the distributions might be a bit different,
+it looks like the UK one is slightly bimodal (it has two peaks),
+whereas the US one is unimodal (it has one peak).
```
ttimbers, line 215: These distributions do not look meaningfully different to me. The bimodalness you are commenting on is an artifact of the binning and low number of observations in the UK data. My conclusion would be that they are fairly similiar, maybe I would play with the binning on another slide to drive this point home to the learners (say bins of 15 or 10).

chapters/en/slides/module3/module3_01.Rmd
```diff
@@ -0,0 +1,274 @@
...
+
+We could explore that data further to find out why this is,
+e.g. maybe the genre is more related to movie length
+and the UK movies in this dataset are of different genres than the US ones?
+
+---
+
+## Comparing multiple distributions grouped by two variables
```
ttimbers, line 223: Is there another country in the data set you could compare? What about China or India? They produce a lot of movies? Which might show more differences to the US movies? Otherwise, I think we need to temper our conclusions on how different these distributions are, as again, I would not support the claim that these differences are meaningful.

chapters/en/slides/module3/module3_01.Rmd
```diff
@@ -0,0 +1,274 @@
...
+
+We can get a rough indication of the genre differences from this plot,
+but as we have mention previously,
+it is not ideal to be comparing marks with different baselines.
+
+Let's unstack the bars to be able to easier compare the histograms.
+
+---
+
+## Layering bars instead of stacking them can make distributions easier to compare
```
ttimbers, line 249: I think the slide title does not go well with this viz. I find it very difficult to actually see these distributions with this plot... A density plot would probably be better, as would a ridge plot. I would actually use this plot to motivate those two other options...

ttimbers (changes requested):
Please see feedback in the review to improve these slides. They are a good start and direction, so nice work on the scope!

chapters/en/slides/module3/module3_01.Rmd
joelostblom: It probably deployed an earlier commit and didn't update

chapters/en/slides/module3/module3_01.Rmd
joelostblom: Yes, the main point I want to get across that it is hard to tell and that we can scale them to make it easier to compare

chapters/en/slides/module3/module3_01.Rmd
joelostblom: Ok will look into this. Would be a good idea to play with binning either way

chapters/en/slides/module3/module3_01.Rmd
joelostblom: Isn't this clear from the text where I say the same thing as your comment? I can add "sometimes" to the title and "(density plots)" after "smooth histograms"?

chapters/en/slides/module3/module3_01.Rmd
joelostblom: I rephrased to say that we can't claim that these differences are meaningful without more data and more plots with different bin sizes. There is unfortunately even fewer movies from other countries in the original dataset and I think none from the two you mentioned. It would be interesting, but I don't think we have time to find that data now.

chapters/en/slides/module3/module3_01.Rmd
joelostblom: I will take out this since it is covered in the previoous modeule

PR #46, opened 2021-01-23
3.2 density plots

joelostblom:


chapters/en/slides/module3/module3_02.Rmd
```diff
@@ -0,0 +1,308 @@
...
+a histogram is not an as unbiased plot as we might think at first,
+especially not if we have few data points,
+where the inclusion or exclusion of just a few points
+makes a big difference for the bar height.
+
+---
+
+## Centering the bins on the data can help create more accurate distribution plots
+
+<img src=/module3/kde-buildup.gif></img>
```
joelostblom, line 53: Ref youtube video that I made this gif from

chapters/en/slides/module3/module3_02.Rmd
```diff
@@ -0,0 +1,308 @@
...
+```{python}
+(alt.Chart(movies).transform_density(
+     'runtime',
+     groupby=['genre', 'country'],
+     as_=['runtime', 'density'])
+ .mark_area().encode(
+     x='runtime',
+     y='density:Q',
+     color='genre')
+ .properties(height=265).facet('country'))
```
joelostblom, line 225: Setting properties instead of changing slide height for now. Tracked in #47

chapters/en/slides/module3/module3_02.Rmd
```diff
@@ -0,0 +1,308 @@
...
+<img src=/module3/kde-buildup.gif></img>
+
+Notes:
+Instead of setting fixed lines along the axis
+and then count plots fully in one bin or another,
+we can create bins that are centered
+on the data and then add them together.
+
+When we center bins on the data,
+we also usually have a round shape
```
ttimbers, line 62: I feel like more needs to be explained here.. Why do we give them a round shape? And it's not really round, its bell-shaped...

chapters/en/slides/module3/module3_02.Rmd
```diff
@@ -0,0 +1,308 @@
...
+Notes:
+We will be using the movies dataset again,
+let's see if we can make a plot
+that is easier to interpret
+than the layered histograms
+from the end of the last slide deck.
+
+---
+
+## Creating a density plot in Altair required two step
```
ttimbers, line 95:
```suggestion
## Creating a density plot in Altair required two steps
```

chapters/en/slides/module3/module3_02.Rmd
```diff
@@ -0,0 +1,308 @@
...
+```{python}
+import pandas as pd
+
+
+movies = pd.read_csv('movies.csv')
+movies
+```
+
+Notes:
+We will be using the movies dataset again,
```
ttimbers, line 87: To answer what question??? We need to provide context & motivation here for what we are doing. For example, 
"We will be using the movies data set again, this time to answer the question what is the minimum and maximum range of runtimes of movies, and what is the most common range of runtimes?

chapters/en/slides/module3/module3_02.Rmd
```diff
@@ -0,0 +1,308 @@
...
+ .mark_area().encode(
+    x='runtime',
+    y='density:Q'))
+```
+
+Notes:
+To visualize a density plot,
+we need to do two things:
+
+1. Calculate the KDE (adding the individual kernels and summing them together)
```
ttimbers, line 112: Don't use KDE with these learners, too much new stuff is going on already, the do not need an acronym to remember as well.

chapters/en/slides/module3/module3_02.Rmd
```diff
@@ -0,0 +1,308 @@
...
+The history genre has the longest running movies
+as well as the most variation
+whereas most animation moves are shorter
+and with less spread within the genre.
+
+There are still a few things we could tweak to improve this plot!
+
+---
+
+## Creating smoother density plots with more calculation steps
```
ttimbers, line 164: I think a slide needs to be devoted to explaining what is going on here, at least at a conceptual level. Maybe an animation like you had in slide 3?

chapters/en/slides/module3/module3_02.Rmd
```diff
@@ -0,0 +1,308 @@
...
+     x='runtime',
+     y='density:Q',
+     color='genre'))
+```
+
+Notes:
+The density areas don't suffer the same issues as the histograms when made transparent.
+The continuous solid shape is easier to follow
+even when it is semi transparent
+overlaps with the other areas.
```
ttimbers, line 207: But eventually it might, if you had A LOT of densities - could you facet vertically to solve such a problem with altair?  Kind of like a ridge plot? If so, it would be nice to add a slide here showing that.

chapters/en/slides/module3/module3_02.Rmd
```diff
@@ -0,0 +1,308 @@
...
+which often creates sharp borders as in this plot.
+
+This is a great default in Altair,
+but many other plotting packages
+instead extend the density beyond the observations in the dataset
+to make it more smooth
+and it is therefore paramount
+to always ask how many observations there are
+before interpreting a density plot.
+
```
ttimbers, line 270: Could adding a mark, like tick marks where each viz goes, be useful here to show that there are only 10 data points, and where they fall? 

This slide is also begging the question, would a histogram be better than a density in this small n situation?

chapters/en/slides/module3/module3_02.Rmd
```diff
@@ -0,0 +1,308 @@
...
+The transparency gives us the advantage
+of knowing that there is not small density area hiding behind a bigger one.
+
+---
+
+## The y-axis does not indicate a meaningful value for density plots
+
+```{python}
+(alt.Chart(movies).transform_density(
```
ttimbers, line 217: Why do you drop steps here? It makes the densities look a lot better and reinforces using it, I would carry it through now.

ttimbers (changes requested):
Some feedback for the densities section!

chapters/en/slides/module3/module3_02.Rmd
joelostblom: I meant to say "rounded" I think. So this part I have simplified based on my understanding of the learners capabilities and previous comments on some section being too difficult, but I am happy to include more detail. In 531 I do mention that there are different kernels and briefly show them but don't have any real discussion around it because "everyone" just uses gaussian kernal for KDEs without any real good reason, it just happens to be the default option. And while gaussian reduces noise compares to a square top had kernel, which makes it looks more aesthically pleasing, I don't know if it is technically "the best" choice". From what I have read, the kernel that best represents the distribution (whatever that really means), is often epanechnikov, but in my testing there is not a large difference vs gaussian so might as well stick with the default. [Some examples here](https://scikit-learn.org/stable/auto_examples/neighbors/plot_kde_1d.html#sphx-glr-auto-examples-neighbors-plot-kde-1d-py
)
 
What do you think would be a suitable level of detail to include here? Since they haven't heard of p-values and error bars, I imagine they don't know of gaussian/normal and mentioning any of these kernels by name is not useful. Something like the suggestion below maybe?

chapters/en/slides/module3/module3_02.Rmd
```diff
@@ -0,0 +1,308 @@
...
+Instead of setting fixed lines along the axis
+and then count plots fully in one bin or another,
+we can create bins that are centered
+on the data and then add them together.
+
+When we center bins on the data,
+we also usually have a round shape 
+instead of a square one,
+which gives the impression of a smoother distribution
+that is more conducive to communicating the shape of the data distribution.
```
joelostblom, line 65:
```suggestion
When we center bins on the data,
we usually give them a rounded shape
instead of a square one.
This removes noise or spikes in the plotted area,
which could arise when using a square kernel,
but it often not informative for us
when trying to get an overall idea of the distribution shape,
and a smoother area is more conducive to communicating the shape of the data distribution.
```

chapters/en/slides/module3/module3_02.Rmd
joelostblom: Good catch, I will switch to something like "adding the rounded bins together"

chapters/en/slides/module3/module3_02.Rmd
joelostblom: Agree that an animation would be great but don't think that I have seen one. I might make a few plots instead. Ideally, I would not talk about this at all. I wish the default value for the steps was higher, I'm going to open a vegalite issue

chapters/en/slides/module3/module3_02.Rmd
joelostblom: You could but[ it is a bit hacky and rather involved](https://altair-viz.github.io/gallery/ridgeline_plot.html). I will see if I can add a simpler example.

chapters/en/slides/module3/module3_02.Rmd
joelostblom: Ideally, I would not talk about this at all. I wish the default value for the steps was higher, I went back to the default to make the code simpler and because we have limited vertical space in the slides

chapters/en/slides/module3/module3_02.Rmd
joelostblom: Good idea, let's add in another mark or add a subplot. One big concern is space on the slide.

I don't know the official answer to the second question, I would highlight that each come with their own caveats but those of the histogram might be more tolerable for a low n, I have a question on this in the lab

chapters/en/slides/module3/module3_02.Rmd
joelostblom: Added it back in to the immediate slide but not when we switch to only having 10 data points.

chapters/en/slides/module3/module3_02.Rmd
joelostblom: I rephrased this section

PR #48, opened 2021-01-26
M2 review

hfboyce:
In this PR: 

- Added consistency in slides.
- Fixed slides spelling, typos, note fields, image and table sizes. 
- Added True/False Qs.
- Added Multiple choice Qs.
- Added coding Qs and files (exc, solution).
- Added Data (already pushed to binder branch).
- Renamed files .
- Changed chapter structure.


ttimbers:
Title header for this module is wrong in this branch: "Module 2: Visual Encoding **X**hannels"

chapters/en/module2.md
```diff
@@ -0,0 +1,789 @@
...
+- Are you specifying `Year` on the x-axis?
+- Are you specifying `color = 'Origin'`?
+- Are you giving the plot a title using `.properties()`?
+
+</codeblock>
+
+
+**Question**      
+
+Which origin's fuel efficiency improved the most between the years 1970 to 1982?
```
ttimbers, line 347: I cannot click these radio buttons.

chapters/en/module2.md
```diff
@@ -0,0 +1,789 @@
...
+
+<opt text="<code>bin=alt.Bin(bins=30)</code>">
+
+You are close but the argument `bins` within the helper function `alt.Bin()` isn't quite right. 
+
+</opt>
+
+<opt text="<code>bin=alt.Bin(maxbins=30)</code>">
+
+Nice!
```
ttimbers, line 489: When you choose this correct answer, it outputs: "Incorrect. Nice!" in red...

chapters/en/module2.md
```diff
@@ -0,0 +1,789 @@
...
+
+
+We are interested in obtaining the number of penguins we collected of each species. 
+
+Tasks: 
+
+- Using a bar plot, visualize the count of penguins for each species in the data. 
+- Choose the appropriate rotation, given the axis labels. 
+- Sort your bars in an appropriate order. 
+- Make sure to give it a title and set the dimensions to a height of 350 and a width of 500.
```
ttimbers, line 525: The height is too high here and violates the ink-space rule of thumb, please drop down to 100? 150?

chapters/en/module2.md
```diff
@@ -0,0 +1,789 @@
...
+**When you see `____` in a coding exercise, replace it with what you assume to be the correct code. Run the code to see if you obtain the desired output
+and submit it to validate if you were correct.**
+
+
+Let's now see what the distribution of penguin flipper length looks like. 
+
+Tasks: 
+
+- Using bars, show the counts of the `flipper_length_mm` values. 
+- Make sure to give it a title and set the dimensions to a height of 350 and a width of 500.
```
ttimbers, line 580: plot is too wide, reduce width to 300?

chapters/en/module2.md
```diff
@@ -0,0 +1,789 @@
...
+
+<exercise id="13" title="Creating Subplots via Faceting" type="slides,video">
+<slides source="module2/module2_13" shot="1" start="0:003" end="07:12"> </slides>
+</exercise>
+
+
+<exercise id="14" title="True or False: Facets">
+
+**True or False**       
+*If faceting, it's important to assign a color channel to the same column.*
```
ttimbers, line 604: I find this question confusing... I need more context...

ttimbers (changes requested):
Comments to improve the q's and fix some issues included,

chapters/en/module2.md
```diff
@@ -0,0 +1,789 @@
...
+
+</choice>
+
+</exercise>
+
+
+<exercise id="6" title='Checking your "Area" of Expertise'>
+
+**Question 1**      
+Before making any plots, what is an important first step in visualizing data?
```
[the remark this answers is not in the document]
joelostblom: Great point about sketching on paper, this is one of the most useful tips I know of myself! Especially in the beginning, it really helps thinking about the data. The first milesone in 532 does this, but it should really be in 531 already.

In the first slide of the course, we have this:

> We often visualize data in order to help us answer a specific question
we have about our dataset,
but it can also help us generate new questions.

I agree that we can have a slide that talks more about this, maybe the first slide before we create some code?I think that would be 1.2, the text could be something like:

> Before creating a visualization, decide why you are making it.
Is there a specific question you are trying to answer,
like comparing the relationship between two dataframe columns?
Or are you creating a plot to help you understand the structure of your data more in general,
such as plotting the distribution of each column?
In either case, it can be *extremely* helpful to draw out your plot with pen and paper first.
This helps you think about if the plot you are creating makes sense
or if there is another plot better suited for the task at hands.
It also makes it easier to write the code afterwards, since you clearly know what you are expecting.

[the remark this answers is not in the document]
joelostblom: Will update question when slide is updated https://github.com/UBC-MDS/exploratory-data-viz/issues/52

chapters/en/module2.md
hfboyce: Ok I've changed it to this: 

> **True or False**       
*If you faceting and assign a color channel to the same column, you will obtain subplots all in different colours*

> Answer:
> Setting a colour channel column to the same column as we are  facetting on,  will change each of the sub plots to different colours.

Still confusing? Should I come up with a new question?

chapters/en/module2.md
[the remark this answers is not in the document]
joelostblom: @hfboyce Did you end up updating this question and or slide? Otherwise I can do it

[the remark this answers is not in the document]
hfboyce: I did not update this question since I wasn't sure what the plan was.

chapters/en/module2.md
joelostblom: I think this is much better, it is not necessarily "important" to do this.

PR #49, opened 2021-01-31
3.3 multiple distributions

joelostblom:
There is some more work to do here, but this should be good for you @hfboyce and also for a review from @ttimbers.


chapters/en/slides/module3/module3_03.Rmd
```diff
@@ -0,0 +1,245 @@
...
+type: slides
+
+# Comparing many distributions
+
+Notes:
+We have seen have we can use facetting for histograms
+and both faceting and coloring for density plots
+to compare multiple distributions against each other.
+So far we only studied these plots with up to three different distributions to compare,
+but what if we had many more?
```
ttimbers, line 25: Let's give an idea of how many more - say 20? 50? 100? That will let them imagine how poorly the viz's we have shown this far will work.

chapters/en/slides/module3/module3_03.Rmd
```diff
@@ -0,0 +1,245 @@
...
+we would arrive at widely different conclusion to what was going on with our data.
+
+This is why we most of the times need to present a richer representation of the data
+such as a histogram or a density plot.
+Since these do not work well for multiple distributions,
+we will use a compromise: the boxplot.
+
+---
+
+## Boxplots show several key statistic from a distribution
```
ttimbers, line 133: It would be nice to recreate this graphic, showing raw data points (jitter), a density and/or histogram and boxplot all lined vertically on top of one another. This would more clearly illustrate what the boxplot is showing. Also, let's label the box plot with percentages as well? Or even just percentiales? They are easier to understand than quartile, in particular, they will have no idea what an interquartile range is.

chapters/en/slides/module3/module3_03.Rmd
```diff
@@ -0,0 +1,245 @@
...
+in this picture they re the min and min and the max (the range) of our data,
+but they can also be 1.5x the interquartile range
+and then any observations outside this are drawn out as individual points
+(this is the case in Altair).
+
+---
+
+## Boxplots can effectively compare multiple distrubtions
+
+```{python}
```
ttimbers, line 154: Can we show the bar chart and the bloxplot here side-by-side and compare and contrast them? I think that would really bring the message home. I think we should do a lot of this in this course, it's quite effective and illustrative. As it is hard to hold both images in your head at once.

chapters/en/slides/module3/module3_03.Rmd
```diff
@@ -0,0 +1,245 @@
...
+we could not tell which of the distributions it came from,
+but looking at the points directly
+we would arrive at widely different conclusion to what was going on with our data.
+
+This is why we most of the times need to present a richer representation of the data
+such as a histogram or a density plot.
+Since these do not work well for multiple distributions,
+we will use a compromise: the boxplot.
+
+---
```
ttimbers, line 131: I think here it's important to remind them when it is appropriate to use a bar plot (e.g., counts, proportions, percentages, etc).

chapters/en/slides/module3/module3_03.Rmd
```diff
@@ -0,0 +1,245 @@
...
+
+<img src=/module3/barplot-hiding-points.png></img>
+
+[Beyond Bar and Line Graphs: Time for a New Data Presentation Paradigm](https://journals.plos.org/plosbiology/article?id=10.1371/journal.pbio.1002128)
+
+Notes:
+This image illustrates how the same barplot (A)
+could have been generated from several different distributions (B, C, D, or E).
+If we just looked at the barplot,
+we could not tell which of the distributions it came from,
```
ttimbers, line 122: Why is this important? It won't be clear to the students. It would be worthwhile here talking about the outlier in C and how that effects the mean, and how the mean (and thus our interpretation) would change if the outlier was removed (or we collected more data, of which very few would expected to be outliers). As well as talking about  D - in that maybe there are really two groups for the black bar group, and so we maybe should rethink our data and divide it into 3 groups?

Finally, for this slide, you should remove the error bars (the learners will have no idea what these are yet) and the statistical tests beneath, as again, they will have no meaning for these learners.

ttimbers (changes requested):
One big picture closing comment to go along with the smaller feedback, I think we are missing a useful viz for comparing distributions, that is particularly effective when you have smaller data sets (10-100 observations per group) with relatively few groups, which is just a jitter strip chart, with the estimate (e.g., mean) highlights. Something like this (without the error bars):

Screen Shot 2021-02-04 at 11 26 35 AM

Also, it would be good to end with a slide that showed the various distribution visualizations that you discussed, with data that was very suitable for each case, and remind them why each was chosen.

chapters/en/slides/module3/module3_03.Rmd
joelostblom: I think we have highlighted why it is important in the second part of this sentence (we would arrive at a different conclusion regarding the data). I agree we can make it clearer exactly *how* the conclusions would change and will add info  in your example, thanks!

Agree to simplifying the image, I was thinking maybe making my own eventually but just pasted this for now. Altough it is CC BY 4.0, so we could just edit.

chapters/en/slides/module3/module3_03.Rmd
joelostblom: Yeah, Hayley also pointed out that we need to explain IQR etc more. I will make a simplified versus of this when revising instead of the one I copied in when making the draft.

chapters/en/slides/module3/module3_03.Rmd
joelostblom: Agree its good to learn by contrast and will try to include this! I just needto figure out the space constraints and the fact that we weren't planning to teach layouts until lecture 7. We could either teach it earlier and make it on a separate slide.

joelostblom (commented):
> One big picture closing comment to go along with the smaller feedback, I think we are missing a useful viz for comparing distributions, that is particularly effective when you have smaller data sets (10-100 observations per group) with relatively few groups, which is just a jitter strip chart, with the estimate (e.g., mean) highlights. Something like this (without the error bars):

100% Agree with this but it is not possible in Altair. Yes, you read that correctly, a "statistical" visualization library that does not support jittered stripplots, swarmplots, or violin plots. This and the lack of dodging were two of the main reasons I was against having Altair as the default viz library in MDS when it was up for discussion a long time ago (but I'm glad it went in anyways because the interactivity grammar is unparalleled). I think these plots are such key visualization tools that it is non-comprehensible that they don't have higher priority to implement in VegaLite. They do finally seem to be on the horizon from a comment I received on their issue tracker.

So what I have shown is a straight stripchart without jitter when we visualize the one dimensional distribution in the beginning. I will link some of the main issues I have with vegalite at a later time this spring and see if we can do some coordinated pushing/upvoting... 

> Also, it would be good to end with a slide that showed the various distribution visualizations that you discussed, with data that was very suitable for each case, and remind them why each was chosen.

Interesting idea, do you mean without the code? Or coding the three of them and laying them out in a concatenated viz?

chapters/en/slides/module3/module3_03.Rmd
joelostblom: Added, not sure how how intuitive it is to introduce the layout options here, but lets try!

chapters/en/slides/module3/module3_03.Rmd
joelostblom: Tracked in #102

PR #81, opened 2021-02-22
Add tests for M2 exercises.

jkim222383:
For the `test_02_12`, I assumed that the `flipper_length_mm` goes in x-axis and `count()` goes in the y-axis. 
Would it be possible to specify that in the instructions as well?

Similarly, for the `test_02_16`, I assumed that the `body_mass_g` goes in x-axis and `count()` goes in the y-axis. 

Please let me know what changes are needed!

### DIFF
--- a/exercises/en/test_02_07a.py
+++ b/exercises/en/test_02_07a.py
@@ -1,0 +1,19 @@
+def test():
+    # Here we can either check objects created in the solution code, or the
+    # string value of the solution, available as __solution__. A helper for
+    # printing formatted messages is available as __msg__. See the testTemplate
+    # in the meta.json for details.
+    # If an assertion fails, the message will be displayed
+
+    assert not fuel_efficiency is None, "Your answer does not exist. Have you passed in the correct variable?"
+    assert type(fuel_efficiency) == type(alt.Chart()), "Your answer is not an altair Chart object. Check to make sure that you have assigned an alt.Chart object to fuel_efficiency."
+    assert fuel_efficiency.data.equals(data.cars()), "Make sure you are using cars() dataset from vega_datasets."
+    assert fuel_efficiency.mark == 'area', "Make sure you are using the area mark type."
+    assert (fuel_efficiency.encoding.x.field in {'Year', 'Year:temporal', 'Year:T'} or 
+            fuel_efficiency.encoding.x.shorthand in {'Year', 'Year:temporal', 'Year:T'}), "Make sure you are using 'Year' as the x-axis encoding."
+    assert (fuel_efficiency.encoding.y.field in {'Miles_per_Gallon', 'Miles_per_Gallon:quantitative', 'Miles_per_Gallon:Q'} or 
+            "Miles_per_Gallon" in fuel_efficiency.encoding.y.shorthand), "Make sure you are using mean of the 'Miles_per_Gallon' as the y-axis encoding."
+    assert ((fuel_efficiency.encoding.y.field in {'Miles_per_Gallon', 'Miles_per_Gallon:quantitative', 'Miles_per_Gallon:Q'} and fuel_efficiency.encoding.y.aggregate == 'mean') or
+            fuel_efficiency.encoding.y.shorthand in {'mean(Miles_per_Gallon)', 'mean(Miles_per_Gallon):quantitative', 'mean(Miles_per_Gallon):Q'}), "You're very close. Make sure that you are using the mean aggregate for the y-axis encoding."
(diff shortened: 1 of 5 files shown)


exercises/en/test_02_07a.py
```diff
@@ -0,0 +1,19 @@
...
+    assert fuel_efficiency.data.equals(data.cars()), "Make sure you are using cars() dataset from vega_datasets."
+    assert fuel_efficiency.mark == 'area', "Make sure you are using the area mark type."
+    assert (fuel_efficiency.encoding.x.field in {'Year', 'Year:temporal', 'Year:T'} or 
+            fuel_efficiency.encoding.x.shorthand in {'Year', 'Year:temporal', 'Year:T'}), "Make sure you are using 'Year' as the x-axis encoding."
+    assert (fuel_efficiency.encoding.y.field in {'Miles_per_Gallon', 'Miles_per_Gallon:quantitative', 'Miles_per_Gallon:Q'} or 
+            "Miles_per_Gallon" in fuel_efficiency.encoding.y.shorthand), "Make sure you are using mean of the 'Miles_per_Gallon' as the y-axis encoding."
+    assert ((fuel_efficiency.encoding.y.field in {'Miles_per_Gallon', 'Miles_per_Gallon:quantitative', 'Miles_per_Gallon:Q'} and fuel_efficiency.encoding.y.aggregate == 'mean') or
+            fuel_efficiency.encoding.y.shorthand in {'mean(Miles_per_Gallon)', 'mean(Miles_per_Gallon):quantitative', 'mean(Miles_per_Gallon):Q'}), "You're very close. Make sure that you are using the mean aggregate for the y-axis encoding."
```
hfboyce, line 17: You may not need these "or" statements here. Can you check? 
Also you can utilize the in __solution__ to see if the student is typing in a certain string.

exercises/en/test_02_11.py
```diff
@@ -0,0 +1,23 @@
...
+    # string value of the solution, available as __solution__. A helper for
+    # printing formatted messages is available as __msg__. See the testTemplate
+    # in the meta.json for details.
+    # If an assertion fails, the message will be displayed
+
+    # Since we haven't started assigning charts to variable names yet,
+    # this might be the better way to test for the first exercise.
+    # Maybe even for later exercises.
+    assert not penguin_bar is None, "Your answer does not exist. Have you passed in the correct variable?"
+    assert type(penguin_bar) == type(alt.Chart()), "Your answer is not an altair Chart object. Check to make sure that you have assigned an alt.Chart object to penguin_bar."
```
hfboyce: Capitalize A in Altair

hfboyce (changes requested):

hfboyce:
This is really good Junghoo. 

Just 2 things for feedback. 
1. Double check the or conditions here, they may not both be needed. It would be a good idea to experiment if the issue we were having was exclusive to Jupyter or not. 
2. You can utilize the `__solution__` object which checks for a specific string in the student's work. This is not mandatory but it may be easier to check for `area` or `line` like this instead of the actual object itself. 
This is all I've got. You work was great!

hfboyce:
Also I like your new picture! 📸

exercises/en/test_02_07a.py
jkim222383: 1. I initially had 
`fuel_efficiency.encoding.x.field == 'Year' or fuel_efficiency.encoding.x.shorthand == 'Year'` 
but realized the test should also accept `'Year:temporal'`

Something like
```
import altair as alt
from vega_datasets import data

cars = data.cars()

fuel_efficiency = alt.Chart(cars).mark_area().encode(
    x='Year:temporal',
    y='mean(Miles_per_Gallon):quantitative').properties(title="Fuel efficiency over time")

print(fuel_efficiency.encoding.x)
print(fuel_efficiency.encoding.y)

fuel_efficiency
```
prints
```
X({
  shorthand: 'Year:temporal'
})
Y({
  shorthand: 'mean(Miles_per_Gallon):quantitative'
})
```
even though the plot generated still meets the requirements.

I thought about just checking for `Year` in the `field`/`shorthand` like 

`'Year' in fuel_efficiency.encoding.x.field or 'Year' in fuel_efficiency.encoding.x.shorthand` 

but depending on how I order the conditions it would return error because Python doesn't like checking whether a `str` is in an object of type `alt.utils.schemapi.Undefined` (because of the weird Altair behaviour, either `field` or `shorthand` is of this `Undefined` type but I don't know what actually determines which one ends up being used). Plus it would accept incorrect answers like `x=mean('Year')`..

I'll think of a more straightforward way to test the encodings but for now I got a little paranoid (:sweat_smile:) and tested all the encodings in this format.

2. Yeah, thank you for reminding me about __solution__ object!

PR #94, opened 2021-03-08
Test HTML figures for mod 1

joelostblom:
I want to confirm that this works with netlify dploy preview


joelostblom:
@hfboyce I am merging this as a text to confirm they build correctly on the live app. Also I increased the slide height to 660px because I was accidentally zoomed in before which somehow made the text appear smaller on the slides... (yeah weird). Let me know if they slides are too big for you now.

PR #96, opened 2021-03-08
M1 hayley cleanup

hfboyce:
Grammarly and consistency.

### DIFF
--- a/chapters/en/slides/module1/module1_03.Rmd
+++ b/chapters/en/slides/module1/module1_03.Rmd
@@ -12,7 +12,6 @@ source('../../../../slide-setup.R')
 # Remember to also manually update the YAML title above
 ```
 
-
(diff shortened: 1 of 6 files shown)


chapters/en/slides/module1/module1_03.Rmd
```diff
@@ -32,15 +31,15 @@ When learning about data visualization, it is helpful to distinguish between the
 
 Notes: 
 
-There is a plethora of visualization packages in Python.
```
joelostblom, line 35: This always confuses me. Isn't it a single "plethora"? Just as I would say "There is a bucket of toys over there" rather than "There are a bucket of toys over there"? English is weird https://englishlessonsbrighton.co.uk/there-is-a-lot-of-vs-there-are-a-lot-of/

PR #99, opened 2021-03-09
M2 hayley cleanup

hfboyce:
I also accidentally pushed to main.... can you double check that everything is ok on here?


joelostblom:
Most of this is already on main, only some changes to an md file that  Iwill regenerate anyways. I will change the height to 650 on my PR

PR #106, opened 2021-03-09
2.4 - Use urldata to deal with large plot files

joelostblom:
@hfboyce Not sure if this will show up ok in the preview, so you can merge if you want to see the built charts.


chapters/en/slides/module2/module2_01.Rmd
```diff
@@ -29,34 +29,6 @@ Notes:
...
-    color='Origin')
-    
-line_plot = alt.Chart(cars).mark_line().encode(
-    x='Year',
-    y='mean(Miles_per_Gallon)',
-    color='Origin')
-    
-point_plot | line_plot 
-```
-
```
hfboyce, line 58: Were you suppose to delete this or was it accidental?

joelostblom: Intentional. This just looks like a section where I was trying out things, it has nothing to do with the notes, which should be on the previous page.

hfboyce: Shouldn't we update the md file then? 
And what do you want instead of the plots then?

joelostblom: I don't think this is in the md file, do you see it there? The notes should go on the previous slide, so there is nothing here.

hfboyce: It currently looks like this from the deploy preview which I'm going to use. 
image

hfboyce: Everything else is find and I'll merge to master.

joelostblom: I'll check that but merge for now.

PR #95, opened 2021-03-08
Wrote tests for M3 content

ecool50:

PR #107, opened 2021-03-09
m3.2 review elijah joel

joelostblom:
@hfboyce and @ecool50 I needed to do some manual editing when rebasing https://github.com/UBC-MDS/exploratory-data-viz/pull/95 on main before merging. I did it on a new branch just so that we still have the old PR incase something went wrong. Could you please review this PR and approve if it looks good?

You don't need to go through all the commits, just the changed files. The test files should be fine, but the exercises I needed to manually resolve, so check those mostly.

close #95


hfboyce (changes requested):
Hi Joel, 

It looks like we are missing the exercise file for question 13. 

image.

joelostblom:
Good catch @hfboyce, added!

hfboyce:
I'm good here now? Should I merge?

PR #125, opened 2021-03-17
corrected grammar and mismatch transcript to data

hfboyce:
Are we ok to merge this? I just found some issues when reading and ran it through grammarly and corrected some things.


joelostblom (approved):
Thanks for fixing these!

PR #127, opened 2021-03-17
Include M4 tests and fix typo in 04_07 exc/solution.

jkim222383:
`M4_amendments` branch still has `solution_04_09.py` file. Should this be removed?

Also, I noticed that on my end, `exc_04_13.py` was showing up as slide 14 in the module. This doesn't seem to have affected anything else though and the exercise/solution/test worked as expected other than the inconsistent numbering.

Please let me know anything that should be fixed, thanks!


jkim222383:
Oh and please let me know if the tests for checking capitalizations for plot titles/legend titles are too strict or too lax.

I'm using `assert not penguin_heatmap.title.islower()` to test if titles have capitalization but it will let e.g. "penguins in Antarctica" pass, even though "Penguins in Antarctica" might be a more preferred format. For now I just let tests pass if there is any capitalized letter in the title but I can change it so it forces the first letter of all plot/legend titles to be capitalized.

exercises/en/test_04_07.py
```diff
@@ -0,0 +1,38 @@
...
+    # printing formatted messages is available as __msg__. See the testTemplate
+    # in the meta.json for details.
+    # If an assertion fails, the message will be displayed
+
+    assert not pokemon_circleplot is None, "Your answer does not exist. Have you passed in the correct variable?"
+    assert type(pokemon_circleplot) == type(alt.Chart()), "Your answer is not an Altair Chart object. Check to make sure that you have assigned an alt.Chart object to pokemon_circleplot."
+    assert pokemon_circleplot.mark == 'circle', "Make sure you are using the 'mark_circle' to generate a circleplot."
+
+    assert (
+        pokemon_circleplot.encoding.x.shorthand in {'gen:ordinal', 'gen:O'} or
```
hfboyce, line 13: What about just `gen` does that not work?

hfboyce (changes requested):

hfboyce:
Ah! Good Catch with `exc_04_13.py` I'll fixed this on the `M4_amendments` branch so if you pull and merge that should fix it on your end. If you could just rename the test file to `test_04_14` that would be great!

exercises/en/test_04_07.py
jkim222383: Only having `gen` makes altair infer the type as `'quantitative'` instead, but `gen` with `type='ordinal'` works. I modified the test to accept it.
Thanks!

PR #142, opened 2021-03-23
6.1 Narratives

joelostblom:
All three PRs ready for review.  I tried spending less time and not perecting so they are maybe a bit rougher than usual. Could you indicate which things you think are must-fixes when you review @ttimbers and I will address those first?


chapters/en/slides/module6/module6_01.Rmd
```diff
@@ -0,0 +1,340 @@
...
+such as the use of other resources
+
+Changing beef for other products would be highly beneficial,
+but how achievable it is 
+and what cost is associated with making the change
+we don't know from the data we have explored here.
+
+Let's continue our story by looking at the same breakdown for Indonesia,
+is beef production the biggest issue there as well?
+
```
ttimbers, line 241:
```suggestion

If we find evidence to support our hypothesis that global beef demand is a
major driver of deforestation in Brazil, then we might want to consider what interventions
could reverse or slow this effect.

The plot in this slide reveals that raising cattle for the purpose of beef production
is one of the most inefficient ways to produce food for human consumption. Armed 
with this knowledge (and the data to back it up), we could then suggest that one
way to decrease deforestation in Brazil may be to decrease the global demand for beef,
while increasing the demand for other protein-rich foods that can be grown in Brazil
in a more efficient way. 

However, when telling a story, it is important to not jump to conclusions or suggest that 
solutions to problems are simpler than what they might really be. So far we have seen that 
beef may be a major contributing factor to deforestation in Brazil, but there are many others 
aspects of the food production chain that we have not considered.  Additionally, decreasing 
the global demand for beef would be a challenging task, and would likely require the cooperation 
of many countries to implement education campaigns, likely along with other measures. In
telling this story, it would be important to also include these uncertainties and potential challenges.

Let’s continue our story, and next ask whether the large deforestation in Indonesia may be
also due to clearing land for pasture.
```

chapters/en/slides/module6/module6_01.Rmd
```diff
@@ -0,0 +1,340 @@
...
+# TODO save wrangled data
+indonesia_drivers = pd.read_csv('data/deforestation-drivers-indonesia.csv').query('2013 > Year > 2003').groupby('Entity').sum().reset_index()
+alt.Chart(indonesia_drivers, title='Palm oil was the leading cause of deforestation between 2003 and 2013 in Indonesia').mark_bar().encode(
+    x='deforestation_hectares',
+    y=alt.Y('Entity', title='', sort='x'))
+```
+
+Notes:
+This slide shows the biggest drivers of deforestation in Indonesia
+between the years 2003 and 2013.
```
ttimbers, line 256: Again, are we sure we know these are drivers?

chapters/en/slides/module6/module6_01.Rmd
```diff
@@ -0,0 +1,340 @@
...
+From this chart alone it surely seems that decreasing palm oil production
+would lead to less deforestation.
+
+But if palm oil production would decrease
+while the world's demand for vegetable oils remains constant,
+another oil crop would need to take its place.
+
+Similar to what we did for beef,
+let's explore how efficient palm oil and its alternatives are
+when it comes to land use versus yield.
```
ttimbers, line 273:
```suggestion
Here, we see that most of the deforested land in Indonesia is used for palm oil
plantations. We see that much deforested land is also used for small scale 
clearings, timber plantations, and grasslands.

Perhaps you have have read, or heard, negative things about palm oil in the past? 
It has been of focus of late in the media, due to its link to tropical deforestation.

From this chart alone, we might want to recommend that decreasing palm oil production
might be a good intervention to reduce deforestation in Indonesia.

However, if this were to be done, we would need another oil crop would need to take its place 
to keep up with the world's demand for vegetable oils.

Let's explore how efficient palm oil and its alternatives are
when it comes to land use versus yield.
```

ttimbers: Note - we've here somehow dropped the narrative/link that global demand for exports leads to deforestation, is palm oil one of Indonesia's big exports? We have a similar gap in the beef story. Perhaps it is there but it might need to be fleshed out?

chapters/en/slides/module6/module6_01.Rmd
```diff
@@ -0,0 +1,340 @@
...
+
+---
+
+## Palm oil is the biggest contributor to deforestation in Indonesia
+
+```{python}
+# TODO save wrangled data
+indonesia_drivers = pd.read_csv('data/deforestation-drivers-indonesia.csv').query('2013 > Year > 2003').groupby('Entity').sum().reset_index()
+alt.Chart(indonesia_drivers, title='Palm oil was the leading cause of deforestation between 2003 and 2013 in Indonesia').mark_bar().encode(
+    x='deforestation_hectares',
```
ttimbers, line 250: Need a better x-axis label

chapters/en/slides/module6/module6_01.Rmd
```diff
@@ -0,0 +1,340 @@
...
+the story of palm oil is more complex that we can show in a single visualization
+and it is not as easy as saying that we should use palm oil
+solely based on the criteria of efficiency.
+
+For example,
+the type of land where the oil crop grows matters.
+Replacing tropical forests containing rich biodiversity
+with palm oil plantages
+has a more severe impact
+than switching to other oil crops on already cultivated land elsewhere in the world.
```
ttimbers, line 306:
```suggestion
It turns out that palm oil is the most efficient use of land
for the amount of oil produced!
For example, if we wanted to replace it with olive oil,
we would need almost 10x the land to do so!

Just as with Brazil and it's use of pasture from deforested land, 
to raise cattle for beef beef, the story of palm oil is more complex 
then we can illustrate in a single visualization. 
```

chapters/en/slides/module6/module6_01.Rmd
```diff
@@ -0,0 +1,340 @@
...
+
+---
+
+## Summary and recap
+
+TODO Not a 100% of what to put in this slide, bullets points of the transcript? Or maybe the bullets points from the first slide again.
```
ttimbers, line 312: I would put those bullet points again, or even them with a storybook image?

ttimbers (changes requested):
Good first draft. 

I have highlighted some small issues with the plots to clean up, and re-wrote some of the narrative (as I will be filming this). There also seemed to be some overstatement of results/conclusions in the narrative. I tried to edit/fix that, but I left some questions in where I need your help @joelostblom to do this.

chapters/en/slides/module6/module6_01.Rmd
joelostblom: We don't have data on how much of each product is exported, that is why I don't talk about it and phrased this part as a second story. I think we can add a sentence to make that clear, partiuclarly now that we are explicitly continuing the same story.

chapters/en/slides/module6/module6_01.Rmd
joelostblom: I think the last part here is important, what is the reason for removing it?

chapters/en/slides/module6/module6_01.Rmd
```diff
@@ -0,0 +1,340 @@
...
+which leads to permanent clearing of forest
+to create pastures for cattle grazing.
+
+There is also a smaller contributions from small scale clearings,
+fires, and commercial crops.
+
+Although beef is the major culprit in rain forest deforestation,
+it could be the case that our demand for more food has simply increased
+and cattle herds are an effective use of land to create food for human consumption.
+Let's flip to the next slide to find out!
```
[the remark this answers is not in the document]
joelostblom: I don't think we say anything that is suggesting the it is global beef demand that drives this. I avoided mentioning global and local because the commodity breakdown does not have export and import numbers, those are only available for the total. I can add an explicit sentence about what we are not saying.

Regarding the pasture use, it is [specifically for beef as per the end of this section](https://ourworldindata.org/drivers-of-deforestation#is-soy-production-driving-deforestation). I have not read those references because I think it is outside what we have time for and that Our world in data is generally correct based on what I have fact checked in the past. I can change the entry to say "pasture for cattle" or something like that.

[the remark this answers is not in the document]
joelostblom: Regarding drivers that is what they are called in the linked article and its references as far as I can see, so I don't think we should call them something else. I agree it would be interesting to read the scientific publication and understand how they determine what is a driver and what is not, but not having time for that I would go with how they are classified by those researchers.

Regarding small scale clearings, I looked it up in the publication and it is used for a variety of things:

> Small-scale clearing was identified by its size and postclearing land use (combination of cropland, pasture, orchards, and residences) for older clearings and by size only for the fresh clearings. Median area of loss patches identified as small-scale clearing is 5 ha. Only 24% of small-scale clearing sample pixels fall within the most recent INCRA (National Institute of Colonization and Agrarian Reform) settlement map, which indicates that these small-scale clearings are created not only by smallholders (rural settlers) but also by agroindustrial enterprises.
>
> https://advances.sciencemag.org/content/3/4/e1601047

I'll see if I can add that but I also don't want to get too technical

chapters/en/slides/module6/module6_01.Rmd
```diff
@@ -0,0 +1,340 @@
...
+Of the deforestation today,
+95% occurs in tropical regions,
+which is particularly severe
+since old rain forests are not easily replaced after they are cut down.
+
+---
+
+## Half of the tropical deforestation is localized to only two countries
+
+```{python}
```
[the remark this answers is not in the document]
joelostblom: The main advantage I see with having them in the same plot is that they are directly comparable and we can see that Brazil and Indonesia has deforestation comparable to entire continents. I thought about removing mexico and India to streamline a bit. I might dothat. I will at least mentioning something about this in the text, but I will label this as "nice-to-have", let me know if you think it is critical.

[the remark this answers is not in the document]
joelostblom: Ok I fixed this to remove the countries I mentioned at least. It would be nice with all countries for each continent but we don't have that data easily accessible.

chapters/en/slides/module6/module6_01.Rmd
joelostblom: I was thinking about using protein per m2 instead of kcal, which might be a more fair comparison. Or I don't know, I think both matters and it is best to show both but not doing that here for now. I'll keep it as is for now but fix at the end if I have time.

chapters/en/slides/module6/module6_01.Rmd
joelostblom: Yes, see the longer reply above.

chapters/en/slides/module6/module6_01.Rmd
joelostblom: Putting the bullets point again in the interest of time

chapters/en/slides/module6/module6_01.Rmd
ttimbers: Because what is being said is quite complex, and kind of goes against palm oil being more efficient - which we just showed in the plot. 

I think the problem is the ordering, now that I reflect more on it. Perhaps we should change it to this:

_It turns out that palm oil is the most efficient use of land
for the amount of oil produced!
For example, if we wanted to replace it with olive oil,
we would need almost 10x the land to do so!

So does it make sense for Indonesia to keep clearing
forests and creating more palm plantations because they
are the most efficient oil crop?

No, our conclusions cannot stop here. Things are not so 
simple. The type of land where the oil crop grows also matters.
Replacing tropical forests containing rich biodiversity
with palm oil plantations has a more severe, negative impact
on the environment compared to changing the type of 
oil crops being grown on land that has already been deforested 
and used for agriculture elsewhere in the world.

Just as with Brazil and it's use of pasture from deforested land, 
to raise cattle for beef beef, the story of palm oil is more complex 
then we can illustrate in a single visualization._

chapters/en/slides/module6/module6_01.Rmd
ttimbers: I agree we should make it explicit we do not have such data, so the link is not there for sure.

joelostblom: I added this to the brazil section already and will mention it here again.

chapters/en/slides/module6/module6_01.Rmd
joelostblom: Sounds great! Adding a mention about sustainble palm oil

PR #144, opened 2021-03-23
6.3 layouts

joelostblom:


chapters/en/slides/module6/module6_03.Rmd
```diff
@@ -0,0 +1,179 @@
...
+
+TODO fix so that this code shows up correctly on the slide
+
+---
+
+## Placing plots next to each other allows us to make more direct comparisons
+
+```{python}
+choropleth | points_on_map
+```
```
joelostblom, line 68: image

chapters/en/slides/module6/module6_03.Rmd
```diff
@@ -0,0 +1,179 @@
...
+but it takes less typing to use the operator.
+
+We can see that the legends are stacked to the right of both plots.
+
+---
+
+## Plots can also be concatenated vertically
+
+```{python}
+choropleth & points_on_map
```
joelostblom, line 85: image

chapters/en/slides/module6/module6_03.Rmd
```diff
@@ -0,0 +1,179 @@
...
+where we get the barplot to show up below both the maps
+as desired.
+
+---
+
+## Titles can be added to each level of the layout
+
+```{python}
+centered_map_title = alt.TitleParams(text='Map comparison', anchor='middle')
+((points_on_map | choropleth) .properties(title=centered_map_title) & bars).properties(title='US population by state')
```
joelostblom, line 163: image

chapters/en/slides/module6/module6_03.Rmd
```diff
@@ -0,0 +1,179 @@
...
+
+We could also have used the `alt.vconcat(choropleth, points_on_map)`,
+but it takes less typing to use the operator.
+
+---
+
+## Vertical and horizontal concatenation can be combined
+
+```{python}
+bars = alt.Chart(state_pop).mark_bar().encode(
```
ttimbers, line 100: Can we plot something other than population for these bars? Otherwise it is redundant? Or perhaps even better, is keep the bars with pop'n, but plot something like asma cases per 1000 on the clorepleth map?

Also, wouldn't it be better to stack the maps vertically, and then have a vertical bar chart so that you can easily read the state names? You might want to combine some states into other to make this work, but I think that would be most effective.

chapters/en/slides/module6/module6_03.Rmd
```diff
@@ -0,0 +1,179 @@
...
+In Altair,
+we can create the figure layout
+by combining the vertical and the horizontal concatenation operators.
+
+In this slide,
+we create a barplot of the population in each state
+that we want to place under the two maps.
+But as we can see in this slide,
+this places the barplot under only the rightmost map.
+Why is that?
```
ttimbers, line 126:
```suggestion
Often we want to create more complex figure layouts
where visualizations are stacked both horizontally and vertically.
Such layouts allow us to present connected, or related information near to one another,
and can be useful if there is a key comparison we want to make
between multiple visualizations. 

Yes, we could always place each visualization on its own slide or page 
(when sharing figures on paper),
however, our working memories are only so good. 
We often forget key details about visualizations we 
just saw when it is removed and we are presented with a new one.

In Altair,
we can create the figure layout
by combining the vertical and the horizontal concatenation operators.

In this slide,
we create a barplot of the population in each state
that we want to place under the two maps.
But as we can see in this slide,
this places the barplot under only the rightmost map.
Why is that?
```

chapters/en/slides/module6/module6_03.Rmd
```diff
@@ -0,0 +1,179 @@
...
+
+We want the horizontal concatenation between the maps 
+to occur first
+and just like with mathematical expression,
+we can use parenthesis to indicate 
+which operations should have priority.
+
+You can see an example of that in this slide,
+where we get the barplot to show up below both the maps
+as desired.
```
ttimbers, line 155: Should we somewhere discuss why this version of the panel figure is more effective? We seem to assume students will just know that, but I don't think we have taught them about the effectiveness of panel figures.

```suggestion
The reason that the plot in the previous slide did not look like we wanted it to, 
was that the `&` operator has higher priority than `|` operator.

The resulted in the choropleth and barplot being vertically concatenated
before the point map was horizontally concatenated.

What we really want is the horizontal concatenation between the maps 
to occur first, followed by the vertical concatenation of the barplot.

Just as we use parentheses in mathematics to change the default order 
of mathematical operations, we can use parentheses to change the default order 
of concatenation operations.

You can see an example of that in this slide,
where we place the parentheses around the `points_on_map | choropleth`
to force the horizontal concatenation between the maps 
to occur first, and that leads us to get the barplot to show up below both the maps
as desired.
```

chapters/en/slides/module6/module6_03.Rmd
ttimbers, line 163: This code line is too long, please break it across multiple lines.

chapters/en/slides/module6/module6_03.Rmd
```diff
@@ -0,0 +1,179 @@
...
+
+---
+
+## Titles can be added to each level of the layout
+
+```{python}
+centered_map_title = alt.TitleParams(text='Map comparison', anchor='middle')
```
ttimbers, line 162: I am hopeful we change what the chlorepleth map is telling us (and now come to thing of it the point map too) so that not all 3 maps are showing the same information. Then we can use a more meaningful title here.

ttimbers (changes requested):
I think we need a somewhat large change here for the data. Our teaching here is not as effective as it could be because we (currently) do not demonstrate/show a case where it is useful to have a panel figure. All our figures show the same data. That was useful in the last slide deck, but I think it not helpful, and actually a bit harmful here. 

Can we plot something like asma cases per 1000 on the clorepleth map, population for the bars, and maybe something like maximum spring pollen count for the point plots? Related information that is useful to have in one panel, that would be difficult to understand as well if we split it up across slides or pages?

PR #157, opened 2021-04-08
7.2 advanced interactivity

joelostblom:
Ready for review, not so great chapter name maybe.


chapters/en/slides/module7/module7_02.Rmd
```diff
@@ -0,0 +1,342 @@
...
+1. Create a new selection that specifies which columns should be used for selections (the `fields` parameter).
+2. Add this new selection to the bar plot
+3. Add an encoding to the scatterplot that depends on the new `click` selections. Here we use `opacity`.
+
+Now that we click the bars
+the opacity of the corresponding points changes
+as we would expect!
+
+Note that we can still click and drag in the scatterplot
+since our `brush` selection is unchanged from previously.
```
ttimbers, line 139:
```suggestion
To link separate charts together,
we need to be explicit about what happens when a selection is made.

For example,
to link a click of a bar 
to a selection in the scatterplot,
we need to say which data frame column should be used for the selection.

In this case, `"Origin"` is the column that we want to use 
for the selection, as it is the common encoding between 
both charts. 

If there were two or more data frame columns that were common
encodings in both charts, then you would select the one 
that made more sense for the question you were asking.

To successfully link this selection
we need to do three things:

1. Create a new selection that specifies which columns should be used for selections (the `fields` parameter).
2. Add this new selection to the bar plot
3. Add an encoding to the scatterplot that depends on the new `click` selections. Here we use `opacity`.

Now that we click the bars
the opacity of the corresponding points changes
as we would expect!

Note that we can still click and drag in the scatterplot
since our `brush` selection is unchanged from before.
```

chapters/en/slides/module7/module7_02.Rmd
```diff
@@ -0,0 +1,342 @@
...
+we could use the legend to make the selections.
+We can specify to Altair that we want to make the legend interactive,
+by binding `selection_multi` to it.
+
+We then need to link this new selection
+(called `click_legend` to distinguish it from the previous slide)
+to the points opacity encoding and add it to the plot.
+
+Now if you click the legend,
+it will select the corresponding observations in the scatterplot!
```
ttimbers, line 167:
```suggestion
We don't always have a bar chart to drive the selections from.
What if we only create a scatterplot,
but still want to be able to select data from different origins?

Instead of using a separate chart,
we could use the legend to make the selections.

We can tell Altair that we want to make the legend interactive,
by binding `selection_multi` to it.

We then need to link this new selection
to the points opacity encoding and add it to the plot.

Now if you click the legend,
it will select the corresponding observations in the scatterplot!
```

chapters/en/slides/module7/module7_02.Rmd
```diff
@@ -0,0 +1,342 @@
...
+
+---
+
+## A selection can be bound to the legend instead of a particular chart
+
+```{python}
+click_legend = alt.selection_multi(fields=['Origin'], bind='legend')
+points = points.encode(opacity=alt.condition(click_legend, alt.value(0.9), alt.value(0.2)))
+
+points.add_selection(click_legend)
```
ttimbers, line 149: Reusing issue again here...

chapters/en/slides/module7/module7_02.Rmd
```diff
@@ -0,0 +1,342 @@
...
+---
+
+## Linking selections from separate charts requires explicit columns to select on
+
+```{python}
+click = alt.selection_multi(fields=['Origin'])
+bars = bars.add_selection(click)
+points = points.encode(opacity=alt.condition(click, alt.value(0.9), alt.value(0.2)))
+
+points & bars
```
ttimbers, line 109: Reusing the scatter plot and barplot code here, and overwriting the points and bar selection is a bit confusing and adds unnecessary cognitive load in my opinion. Let's just make a fresh scatter plot and bar chart from scratch each time please.

I know space might be an issue, but if you do horizontal concatenation, you might have enough space?

chapters/en/slides/module7/module7_02.Rmd
```diff
@@ -0,0 +1,342 @@
...
+Now if you click the legend,
+it will select the corresponding observations in the scatterplot!
+
+---
+
+## Legend selections can be added to multiple charts
+
+```{python}
+bars = bars.encode(opacity=alt.condition(click_legend, alt.value(0.9), alt.value(0.2)))
+(points & bars).add_selection(click_legend)
```
ttimbers, line 175: Reusing issue again here...

chapters/en/slides/module7/module7_02.Rmd
```diff
@@ -0,0 +1,342 @@
...
+Ideally only the bars would change length,
+but the axis would stay the same.
+
+In this slide we calculate the max number of cars
+from any region
+and then use this number to fix the extent of the x-axis.
+Now that we select in the scatterplot,
+only the lengths of the bars changes,
+but the range of the axis remains the same.
+
```
ttimbers, line 249: What about fixing the number of categories on the y-axis? Should we do that too? It might make it easier, especially in plots with more than 3 categories...

chapters/en/slides/module7/module7_02.Rmd
```diff
@@ -0,0 +1,342 @@
...
+In addition to adding one more visualization for the asthma cases,
+using the map to identify states can be more helpful
+than just adding a tooltip to the scatterplot
+since we might already have an intuition 
+for where different states are located on the map
+and don't have to go through each of the point's tooltip
+looking for the state name of interest.
+
+TODO we could change this to be another variable in the map like population, I don't have a strong opinion, both have advantages.
+
```
ttimbers, line 337: This slide kind of begs the linked view to also go in the other direction - from the scatter plot to the map. For example, it would be cool to have the state name show up when hovering over a point, and then have that state highlighted on the map. Also, it would be cool to be able to select, or brush over multiple points and see them highlighted on the map.

chapters/en/slides/module7/module7_02.Rmd
```diff
@@ -0,0 +1,342 @@
...
+We could do this one by one,
+or both at once as we do in this slide.
+
+---
+
+## Instead of highlighting, a selection can be used for filtering data
+
+```{python}
+bars = bars.transform_filter(brush)
+(points & bars).add_selection(click_legend)
```
ttimbers, line 193: Reusing issue again here...

chapters/en/slides/module7/module7_02.Rmd
```diff
@@ -0,0 +1,342 @@
...
+neat!
+
+---
+
+## Fixing the extent of the x-axis makes the bars easier to compare
+
+```{python}
+bar_chart_max = cars.groupby('Origin').size().max()
+bars = bars.encode(alt.X('count()', scale=alt.Scale(domain=[0, bar_chart_max])))
+(points & bars).add_selection(click_legend)
```
ttimbers, line 232: Reusing issue again here...

ttimbers (changes requested):
Good first draft again! I have gone through and done a high-level review of the entire deck. I have only had time to carefully edit the transcript for the first 5 slides so far. Will do the last 5 transcript edits tomorrow.

chapters/en/slides/module7/module7_02.Rmd
```diff
@@ -0,0 +1,342 @@
...
+
+In addition to adding one more visualization for the asthma cases,
+using the map to identify states can be more helpful
+than just adding a tooltip to the scatterplot
+since we might already have an intuition 
+for where different states are located on the map
+and don't have to go through each of the point's tooltip
+looking for the state name of interest.
+
+TODO we could change this to be another variable in the map like population, I don't have a strong opinion, both have advantages.
```
ttimbers, line 336:
```suggestion
To link an interaction on a map to another chart,
we use the same logic as we did in previous slides
for linking the bar chart and scatterplot.

Here we create a separate scatterplot
where the stroke colour (outline) of each points depends on
whether we have selected that state on the map or not.

By default,
all states on the map are selected 
so all the points have a black outline.
As we hover over a state in the map,
the outline changes to white for the points
whose states are not selected in the map.

Using the map to identify states can be more helpful
than just adding a tooltip to the scatterplot
since we might already have an intuition 
for where different states are located on the map
and don't have to go through each of the point's tooltip
looking for the state name of interest.

TODO we could change this to be another variable in the map like population, I don't have a strong opinion, both have advantages.
```

ttimbers (changes requested):
Done reviewing the transcript for this slide deck!

chapters/en/slides/module7/module7_02.Rmd
```diff
@@ -0,0 +1,342 @@
...
+In contrast to the two scatterplots in an earlier slide,
+these charts are not created from the same base chart,
+so there selections are not linked.
+
+What if we want to link the selections
+so that when we select a region of origin in the bar chart
+the points in the scatter plot that are from the same origin would be highlighted?
+
+This would be useful to study observations from each Origin separately,
+so how can we perform this linkage in Altair?
```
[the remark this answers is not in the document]
joelostblom: Yeah, I wasn't sure if this was correct when I wrote it but didn't have time to test fully. I did now and the behavior occurs when when there is an aggregate chart create. If I remember correctly about how VL works, it creates a new dataframe with the aggregate data and there is no link to the original data, so automatic linking is not possible.

chapters/en/slides/module7/module7_02.Rmd
```diff
@@ -0,0 +1,342 @@
...
+```
+
+Notes:
+When creating a combined charts the syntax 
+for adding the legend interactivity is similar,
+but we need to add the selection to both charts
+since the legend belongs to both of them.
+
+We could do this one by one,
+or both at once as we do in this slide.
```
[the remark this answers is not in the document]
joelostblom: less typing if both

chapters/en/slides/module7/module7_02.Rmd
```diff
@@ -0,0 +1,342 @@
...
+This type of interaction is sometimes called a dynamic query.
+
+We're keeping the legend here since we have conditions 
+in the charts' encodings that depend on the `legend_click` selection variable,
+but we could remove it if we change these.
+
+Now that we have select points by dragging in the scatter plot
+we can see that the barchart updates the count
+to reflect the number of points we have selected of each color,
+neat!
```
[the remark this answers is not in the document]
joelostblom: Unfortunately we need to keep it or recreate the chart. All interactions that occur inside condition of a chart need to be added as selections for the VL specification to be valid.

chapters/en/slides/module7/module7_02.Rmd
joelostblom: I don't think this is trivial,otherwise would be good.

chapters/en/slides/module7/module7_02.Rmd
joelostblom: adding to wishlist

chapters/en/slides/module7/module7_02.Rmd
```diff
@@ -0,0 +1,342 @@
...
+2. Add this new selection to the bar plot
+3. Add an encoding to the scatterplot that depends on the new `click` selections. Here we use `opacity`.
+
+Now that we click the bars
+the opacity of the corresponding points changes
+as we would expect!
+
+Note that we can still click and drag in the scatterplot
+since our `brush` selection is unchanged from previously.
+
```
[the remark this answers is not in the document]
joelostblom: We show this with transform filter a few slides later. The order could be switch to show that first but I think it is also natural to talk about how to link clicks to a legend instead of a chart so keeping as is.

PR #158, opened 2021-04-09
7.3 widgets

joelostblom:


chapters/en/slides/module7/module7_03.md
```diff
@@ -0,0 +1,389 @@
...
+Here we’re adding a set of radio buttons for the maturity rating of the
+movies and a dropdown button for the genre selection.
+
+We create the radio buttons and dropdown separately, and then add then
+bind them both to the `selection_single` object to their respective
+fields.
+
+To trigger a selection, we now need to both click a radio button **and**
+make a selection in the dropdown.
+
```
ttimbers, line 174: What about checkboxes? Radio buttons are useful, if you are sure you might not want to filter > 1 category... but there are cases I can imagine in the plot above where we might want to filter  > 1 rating category. Can we easily do this in altair? If so, I think we should show this.

(note: if this takes too much time, please move this to the wishlist)

chapters/en/slides/module7/module7_03.md
```diff
@@ -0,0 +1,389 @@
...
+``` python
+slider = alt.binding_range(name='Tomatometer ')
+select_rating = alt.selection_single(
+    fields=['Rotten_Tomatoes_Rating'],
+    bind=slider)
+
+points.encode(
+    opacity=alt.condition(select_rating, alt.value(0.7), alt.value(0.1))
+).add_selection(select_rating)
```
ttimbers, line 187: Should we go back to full code here to remind folks where we are, as we have now dropped the radio buttons and legend selections? And should we keep the legend selections here? As again, there are a lot of categories...

chapters/en/slides/module7/module7_03.md
```diff
@@ -0,0 +1,389 @@
...
+and we do it via `binding_range()`. Here we link the slider to the same
+column as on the x-axis, so that you can see how it works.
+
+If you drag the slider around, you can see that the default behavior is
+to only highlight points that are of the exact values as the slider.
+
+This is useful for selection widgets like the dropdown, but for the
+slider we want to make comparisons such as bigger and smaller than.
+
+Let’s see how to do that in the next slide.
```
ttimbers, line 207:
```suggestion
Notes: So far we have seen how to select categorical values from
dropdown menus and radio buttons. To instead highlight based on
quantitative values, we can use a slider.

Adding a slider is similar to adding a dropdown or set of radio buttons,
and we do it via `binding_range()`. Here we link the slider to the same
column that is mapped to the x-axis, so that you can see how it works.

If you drag the slider around, you can see that the default behaviour is
to only highlight points that share the x-column values that are of the 
exact values as the slider - this is because they are both mapped to the 
`Rotten_Tomatoes_Rating` column.

Now that we understand how we can create a slider, let's next see how to  
use it in a slightly more realistic setting, such as highlighting values above,
or below a threshold.
```

chapters/en/slides/module7/module7_03.md
```diff
@@ -0,0 +1,389 @@
...
+---
+
+## Highlighting points smaller or bigger than a slider value
+
+``` python
+points.encode(
+    opacity=alt.condition(
+        alt.datum.Rotten_Tomatoes_Rating < select_rating.Rotten_Tomatoes_Rating,
+        alt.value(0.7), alt.value(0.1))
+).add_selection(select_rating)
```
ttimbers, line 218: Do we want to show more code than this if we have space?

chapters/en/slides/module7/module7_03.md
```diff
@@ -0,0 +1,389 @@
...
+
+Now that we drag the slider you can see that all the points are
+highlighted that have a value less than the slider value.
+
+---
+
+## Customizing a slider widget
+
+``` python
+slider = alt.binding_range(name='Tomatometer ', min=10, max=60, step=10)
```
ttimbers, line 247: Why does the slider only go up until 60?

And again, do we want to show more code than this if we have space?

chapters/en/slides/module7/module7_03.md
```diff
@@ -0,0 +1,389 @@
...
+compare against the column with the Rotten Tomatoes Ratings, we use
+`alt.datum.Rotten_Tomatoes_Rating`.
+
+This might seem cumbersome at first, but it is necessary to avoid
+ambiguity with only writing `"Rotten_Tomatoes_Rating"`, which in
+comparison expressions means just the string and is not a reference to a
+column in the dataframe.
+
+Now that we drag the slider you can see that all the points are
+highlighted that have a value less than the slider value.
```
ttimbers, line 240:
```suggestion
Notes: 
Let's say that we want to create a chart, where we highlight values
which are smaller than the value specified by the slider. To do this
we need to compare the current value of the slider with the observations'
values held by the column we are filtering on.

The current value of the slider can be accessed via
`select_rating.Rotten_Tomatoes_Rating`, and Altair has a special `datum`
object to access values from different columns in the dataframe. To
compare against the column with the Rotten Tomatoes Ratings, we use
`alt.datum.Rotten_Tomatoes_Rating`.

This might seem cumbersome at first, but it is necessary to avoid
ambiguity with only writing `"Rotten_Tomatoes_Rating"`, which in
comparison expressions means just the string and is not a reference to a
column in the dataframe.

Now that we drag the slider you can see that all the points are
highlighted that have a value less than the slider value.
```

chapters/en/slides/module7/module7_03.md
```diff
@@ -0,0 +1,389 @@
...
+</iframe>
+
+Notes: Just as with dropdowns we can set an explicit start value to have
+an active selection when the chart first appears.
+
+For slider widgets we can also define the range and step size for the
+slider as we have done in this slide.
+
+Now that we drag the slider handle it jumps ten steps at a time, instead
+of the default single step.
```
ttimbers, line 270: The last sentence begs the question, what is the default step size?

```suggestion
Notes: Just as with dropdowns we can set an explicit start value to have
an active selection when the chart first appears. Here we have set the 
chart so that the slider will be set at 20 when the chart first loads via
setting `init={'Rotten_Tomatoes_Rating': 20})` in `alt.selection_single`.

For slider widgets we can also define the range and step size for the
slider as we have done here, setting the minimum to be 10, the maximum
to be 60, and a step size of 10.

Now that we drag the slider handle it jumps ten steps at a time, instead
of the default single step.
```

chapters/en/slides/module7/module7_03.md
```diff
@@ -0,0 +1,389 @@
...
+
+---
+
+## It is more useful to bind a slider to a dataframe column not displayed on the chart axes
```
ttimbers, line 274: I am not sure about this slide... It's quite unsatisfying, as we don't actually ever see a slider being used effectively... Can we try something else before this, where the slider is effective? Like worldwide gross? And then come here and see that it doesn't work very well for temporal data? 

Also, can you please more full explain why it doesn't work well for temporal data? It's left here currently as a bit of a mystery...

chapters/en/slides/module7/module7_03.md
```diff
@@ -0,0 +1,389 @@
...
+
+In addition to showing the distribution of the data this also allows us
+to select a range of values, which is not possible with sliders in
+Altair. We can then drag the selection with the mouse, just as when we
+were selecting in the scatter plots in the first slide deck of this
+module.
+
+If we wanted to we could also style the histogram chart to look more
+like a widget, and remove the incorrect data point, but we won’t do that
+here.
```
ttimbers, line 346:
```suggestion
Notes: In our temporal data case, where we want to filter by year, 
instead of having a a slider, it is more effective to use a
histogram that we can use for selecting a range of years, because
this also gives us information about how many movies we have 
data for each year.

We can use what we learned in the last slide deck to link the selection
between the two plots. The biggest difference here is that instead of a
single or multi selection, we will use the `selection_interval` function
to create a range selection.

In this histogram we can see that there are just a few movies from the
1980s in the data and only one in 2040, which must be an error in the
dataset.

In addition to showing the distribution of the data this also allows us
to select a range of values, which is not possible with sliders in
Altair. We can then drag the selection with the mouse, just as when we
were selecting in the scatter plots in the first slide deck of this
module.

If we wanted to we could also style the histogram chart to look more
like a widget, and remove the incorrect data point, but we won’t do that
here.
```

chapters/en/slides/module7/module7_03.md
```diff
@@ -0,0 +1,389 @@
...
+
+---
+
+## Interval selection can also be used to create a “minimap” for chart navigation
```
ttimbers, line 350: I feel like this would be more effective in something like a chloropleth map? Do you agree? If so, could we swap out the example so that we show that? Or, add a slide after this, showing that?

ttimbers (changes requested):
Great first draft. Hopefully this feedback is helpful for improving it further, and not to much trouble to address.

chapters/en/slides/module7/module7_03.md
joelostblom: Unfortunately multiselect is only possible on another chart/legend, not in widgets. This needs to be added to Vega itself, I have pinged them about it in the past, but does not seem to be a priority.

chapters/en/slides/module7/module7_03.md
joelostblom: I think both could be effective. It would look a bit better if we had more years here. Unfortunately maps in Vega-Lite do not yet work with interactive zooming or panning.

chapters/en/slides/module7/module7_03.md
joelostblom: I think a histogram is almost always better that a slider but added a slide. I would have to look into why date does not work, so putting it on the wishlist

chapters/en/slides/module7/module7_03.md
```diff
@@ -0,0 +1,389 @@
...
+``` python
+mpaa_rating = sorted(movies['MPAA_Rating'].unique())
+dropdown_mpaa = alt.binding_radio(name='MPAA Rating ', options=mpaa_rating)
+dropdown_genre = alt.binding_select(name='Genre ', options=genres)
+select_genre_and_mpaa = alt.selection_single(
+    fields=['Major_Genre', 'MPAA_Rating'],
+    bind={'Major_Genre': dropdown_genre, 'MPAA_Rating': dropdown_mpaa})
+
+points.add_selection(select_genre_and_mpaa).encode(
+    opacity=alt.condition(select_genre_and_mpaa, alt.value(0.7), alt.value(0.1)))
```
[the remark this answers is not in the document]
joelostblom: I'll put this on wish list for now and revisit when I am down with all other changes. In general I like showing all the code if there is room, but also see value in being consistent between the slide and highlighting what is new.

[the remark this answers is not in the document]
joelostblom: I tried a bit with this now, but I can't find a good way of reusing all the code. The code that would be reused is also something they have seen multiple times now: x, y, color, so they should be very used to that. One slide deck has been spent on condition already, and we do show when we change that. I also think the amount of code if we reuse will be confusing and it will be hard to see what is new, especially when flipping through slides on their own.

Issue #163, opened 2021-04-19
Module 7.4 interactive question draft


ttimbers:
##  Q1: True or False

It is easier to save Altair charts as a `.png` or `.svg` image file than as a plain text data file (e.g., `.html` or `.json` file).

Options:
- True
- False

Correct answer: False

> Message for correct (False): Great work! Given how Altair charts are constructed, it's actually easier to save them as  plain text data files!
> Message for incorrect (True): Try again! Go back to the slides and review how Altair charts are constructed.

## Q2: True of False

The SVG image file format stores data as text with mathematical formulas and is good for photos and complex illustrations.

Options:
- True
- False

Correct answer: False

> Message for correct (False): Great work! SVG image files do store data as text with mathematical formulas, however this means that they are really not great for photos or complex illustrations! 
> Message for incorrect (True): Try again! SVG image files are really not great for photos or complex illustrations! 

##  Q3: MCQ

In Altair, a chart like the one below can be saved as an image file via which of the following ways?

<img width="568" alt="Screen Shot 2021-04-19 at 12 34 32 PM" src="https://user-images.githubusercontent.com/6865016/115303152-1e9fac80-a118-11eb-823a-dfb686e13ee8.png">

A. Using Altair's `alt.save` method (e.g., `alt.save(chart_name, "my-chart.png", "PNG")`)

B. By clicking the three dots and selecting "Save as SVG" or "Save as PNG"

C. Using the `altair_saver` method (e.g., `chart_name.save('my-chart.png')`)

D. A & B

E. B & C

> Correct answer: E
> Message for E: Nice work! There is a way to do this both interactively and programmatically!

> Message for A: Try again! You can save Altair charts programmatically, but not with this syntax!

> Message for B: Try again! Yes, you can save Altair charts interactively, but it's not the only way to save them!

> Message for C: Try again! Yes, you can save Altair charts programmatically, but it's not the only way to save them!

> Message for D: Try again! There is a way to do this both interactively and programmatically, but not with the syntax `alt.save`.

##  Q4: MCQ

Visualizing larger data in Altair takes care and can be challenging because:

A. Data is included in that Altair chart specifications

B. Visualizing larger data is always challenging

C. Altair charts can only support 5000 observations in a single chart

> Correct answer: A
> Message for B: Nice work! And because data is included in that Altair chart specifications, bigger data means bigger chart objects!

> Message for B: Try again! We're asking about what makes this specifically challenging in Altair.

[...]

Issue #162, opened 2021-04-19
Module 7.3 interactive question draft


ttimbers:
##  Q1: True of False

If we create a scatter plot of two quantitative variables, but also have 10, or more, categories that we would like to visualize this for, facetting or using interactive selections can make such charts more effective compared to colouring the points by category?

Options:
- True
- False

Correct answer: True

> Message for correct (True): Great work! Indeed 10, or more, categories are far too many for us to effectively use colour along to differentiate points on a scatter plot. Instead interactivity or faceting are great ways to improve the effectiveness of such a visualization!

> Message for incorrect (False): Try again! In the previous slides we showed you how to do both of these things!

##  Q2: True of False

In Altair, sliders can be used to concurrently set an upper and lower threshold to select a range of observations to highlight.

Options:
- True 
- False

Correct answer: False

> Message for correct (False): Great work! This is not yet possible in Altair. However, instead of using a slider to do this, we can instead use a selection interval from another chart to do this.

> Message for incorrect (True): Not quite! Sadly this is not yet possible in Altair.

##  Q3: Coding

**Instructions:**
Be patient when running a coding exercise for the first time, it can take a few minutes.

**When you see `____` in a coding exercise, replace it with what you assume to be the correct code. Run the code to see if you obtain the desired output and submit it to validate if you were correct.**

Let’s now look at the [penguins](https://www.kaggle.com/parulpandey/palmer-archipelago-antarctica-penguin-data) data set again and see whether how Penguins of each species were studied on each island.

Let’s take a look!

```
import pandas as pd
​
​
penguins_df = pd.read_csv('data/penguins.csv')
penguins_df.info()
```

Create a scatter plot from the Penguins data set that visualizes  culmen depth (y-axis) vs culmen length (x-axis), and name it `slider_scatter`. Colour the points by the species. Add a slider that allows you to set a threshold so that you can highlight the observations where the penguins body mass (in grams) is under the threshold specified by the slider. 

Fill in the blanks in the code below so that the following gets accomplished.

```
import altair as alt
import pandas as pd
​
​
penguins_df = pd.read_csv('data/penguins.csv')

slider = alt.binding_range(name='Body mass (g) ', min=1, max=max(penguins_df.body_mass_g))
select_rating = alt.selection_single(
    fields=['body_mass_g'],
    bind=slider)

slider_scatter = (alt.Chart(penguins_df).mark_circle().encode(
[...]

Issue #161, opened 2021-04-16
Module 7.2 interactive question draft


ttimbers:
##  Q1: True of False

In Altair, click selections can be bound to another chart or the legend.

Options:
- True
- False

Correct answer: True

> Message for correct (True): Great work! This is a great feature of Altair!

> Message for incorrect (False): Try again! In the previous slides we showed you how to do both of these things!

##  Q2: True of False

Interactive features in Altair are only possible for scatter plots and bar charts.

Options:
- True 
- False

Correct answer: False

> Message for correct (False): Great work! You can use Altair to create interactive features with any chart you can make in Altair!

> Message for incorrect (True): Are you sure? What about the interactive maps we saw in the last few slides?

##  Q3: Coding

**Instructions:**
Be patient when running a coding exercise for the first time, it can take a few minutes.

**When you see `____` in a coding exercise, replace it with what you assume to be the correct code. Run the code to see if you obtain the desired output and submit it to validate if you were correct.**

Let’s now look at the [penguins](https://www.kaggle.com/parulpandey/palmer-archipelago-antarctica-penguin-data) data set again and see whether how Penguins of each species were studied on each island.

Let’s take a look!

```
import pandas as pd
​
​
penguins_df = pd.read_csv('data/penguins.csv')
penguins_df.info()
```

Create a grouped, horizontal bar chart where the number of penguins on each island is plotted for each species. You are aiming to have a plot that looks like you have 3 subplots, where each subplot is for each species. Add interactivity, so that you can select which island to highlight on the bar chart by clicking the legend. 

Fill in the blanks in the code below so that the following gets accomplished.

```
import altair as alt
import pandas as pd
​
​
penguins_df = pd.read_csv('data/penguins.csv')

click = alt.selection_multi(fields=____, bind=____)

click_legend = (alt.Chart(penguins_df).mark_bar().encode(
    alt.X('count()', title='Number of Penguins'),
    alt.Y('island', title=None, sort='x'),
    alt.Color('island', title="Island"),
    alt.Row('species', title=None),
    opacity=alt.condition(____, ____))
.add_selection(____)).properties(width=300)
```

#### Solution:
```
import altair as alt
import pandas as pd
​
​
penguins_df = pd.read_csv('data/penguins.csv')

brush = alt.selection_interval()

click = alt.selection_multi(fields=['island'], bind='legend')

click_legend = (alt.Chart(penguins_df).mark_bar().encode(
[...]

Issue #160, opened 2021-04-16
Module 7.1 interactive question draft


ttimbers:
##  Q1: True of False

Panning, zooming and brush selections can be simply added to Altair charts by using the `interactive` method.

Options:
- True 
- False

Correct answer: False

> Message for correct (False): Great work! The `interactive` method 

> Message for incorrect (True): Try again! Brush selections are added via `add_selection` and `alt.selection_interval`, not the `interactive` method.

##  Q2: True of False

In Altair, brush selections are automatically linked between scatter plots that are created from the same dataframe.

Options:
- True
- False

Correct answer: True

> Message for correct (True): Great work! This is a great feature of Altair!

> Message for incorrect (False): Try again! It's easier than you think to link selections between scatter plots in Altair!

##  Q3: Coding

**Instructions:**
Be patient when running a coding exercise for the first time, it can take a few minutes.

**When you see `____` in a coding exercise, replace it with what you assume to be the correct code. Run the code to see if you obtain the desired output and submit it to validate if you were correct.**

Let’s now look at the [penguins](https://www.kaggle.com/parulpandey/palmer-archipelago-antarctica-penguin-data) data set again and see whether species differences in body shape also relates to species differences in culmen shape. 

see which penguins are more abundant. Are there penguins that have a greater flipper length (flipper_length_mm) but with a smaller body mass (body_mass_g)? Does the mass of a penguin have a relationship to its flipper length?

Let’s take a look!

```
import pandas as pd
​
​
penguins_df = pd.read_csv('data/penguins.csv')
penguins_df.info()
```

Create two scatter plots from the Penguins data set that have a linked brush selection between them. The first scatter plot should show body mass (y-axis) vs flipper length (x-axis) and be named `linked_scatter`. Colour the points be species. The second plot should show culmen depth (y-axis) vs culmen length (x-axis) - this second scatter plot can be created by just overwriting the encoding of the first. Horizontally concatenate these two scatter plots.

Fill in the blanks in the code below so that the following gets accomplished.

```
import altair as alt
import pandas as pd
​
​
penguins_df = pd.read_csv('data/penguins.csv')

brush = ____

linked_scatter = (alt.Chart(____).mark_circle().encode(
    alt.X(____, title=____, scale=alt.Scale(zero=False)),
    alt.Y(____, title=____, scale=alt.Scale(zero=False)),
    color=alt.condition(____, 'species', alt.value('lightgray')))
    .add_selection(____))
    
[...]

PR #164, opened 2021-04-19
Module7 review

hfboyce:
@joel, I did not change any files besides `module7.md` is this safe to push? 

They are @ttimbers coding and MC questions for M7. 

As per
-  #160
- #161 
- #162 
- #163


hfboyce:
I see no conflict?

PR #176, opened 2025-02-28
Fix slide rendering

joelostblom:
This is a start for how to render slides dynamically, directly from the qmd files without commiting html artifacts, a robust transcript solution via speaker notes, and slide transitions.


joelostblom:
Merging to move on to general styling in another PR

PR #220, opened 2025-07-29
Update module6-02-true_or_false_once_upon_a_time.qmd

matthewconnell:
Fixing typo in Module 6 that caused a question to have no correct answer


joelostblom (approved):
Thank you!

